PR #17, opened 2019-07-25
Cmdtool

ChriMarMe:
Add command line arguments 
-t For choosing multiple tests or ranges of tests
-v Shows version, Copywrite and Licence
-i Forces the tests to not abord on the first error
-l lists all tests with their number
-h for help

### DIFF
--- a/cmd/txt-suite/cmd.go
+++ b/cmd/txt-suite/cmd.go
@@ -1,0 +1,30 @@
+package main
+
+import (
+	"flag"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var testno = flag.String("t", "", "Select test number 1 - 50. e.g.: -t 1,2,3,4,...")
+
+func flagUsed() bool {
+	return testno != nil
+}
+
+func deconstructFlag() ([]int, error) {
+	var testnos []int
+	tmpstrings := strings.Split(*testno, ",")
+	for _, item := range tmpstrings {
+		tmpno, err := strconv.Atoi(item)
+		if err != nil {
+			return nil, err
+		}
+		testnos = append(testnos, tmpno)
+	}
+	//Sort array
+	sort.Ints(testnos)
+	return testnos, nil
+
+}
--- a/cmd/txt-suite/cmd.go
+++ b/cmd/txt-suite/cmd.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"bufio"
 	"flag"
 	"fmt"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


flanfly:
The `-v` option doesn't work for me.

```bash
cave02:txt-suite kai$ ./txt-suite -v
Can't open license file
```

Also, it's copyright, not Copywrite.

Please document that `-t` accepts ranges. Also, it should say that **tests** are ordered by their number, not the numbers itself.

Please don't capitalize "test" in `-l` and replace "Help!" with the name of the tool and it's usage. Something like this:

```bash
Intel TXT test suite.
Usage: txt-suite [-l] [-h] [-v] [-t TESTSPEC]
(detailed docs about command switches)

File: cmd/txt-suite/cmd.go [hunk 1]
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
+	"flag"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var testno = flag.String("t", "", "Select test number 1 - 50. e.g.: -t 1,2,3,4,...")
```
siro20: Should be dynamic depending on available tests

cmd/txt-suite/cmd.go
```diff
@@ -2,17 +2,50 @@ package main
...
 	"sort"
 	"strconv"
 	"strings"
+
+	"github.com/9elements/txt-suite/pkg/test"
 )
 
-var testno = flag.String("t", "", "Select test number 1 - 50. e.g.: -t 1,2,3,4,...")
+var testno = flag.String("t", "", "Select test number 1 - 50. e.g.: -t=1,2,3,4,...")
+var keeprunning = flag.Bool("i", false, "Errors will not stop the testing. Tests will keep going.")
```
siro20: What does -i ?
If there's a fatal error we have to stop/crash.
If there's no fatal error we always keep going.

cmd/txt-suite/cmd.go
```diff
@@ -23,12 +26,31 @@ func stayAlive() bool {
 	return *keeprunning
 }
 
+func showVersion() {
```
siro20: just wrong

- relative path
- a secondary file
- no prefix before Version makes it harder to parse it using scripting languages
- arbitrary text before Version makes it harder to parse it using scripting languages

cmd/txt-suite/cmd.go
```diff
@@ -2,17 +2,50 @@ package main
...
+	fmt.Println("-h : Shows this help")
+	fmt.Println("-l : Lists all Tests with its Test number")
+}
+
+func listTests() {
+	for count, item := range test.TestsCPU {
+		fmt.Printf("Test No: %v, %v\n", count+1, item.Name)
+	}
+	for count, item := range test.TestsTPM {
+		fmt.Printf("Test No: %v, %v\n", count+len(test.TestsCPU)+1, item.Name)
```
PatrickRudolph: use variable outside of for scope

cmd/txt-suite/main.go
```diff
@@ -1,38 +1,67 @@
...
 
+var (
+	testnos []int
+)
+
 func run() {
-	for _, t := range test.TestsCPU {
-		if !t.Run() && t.Required {
-			return
+	if len(testnos) != 0 {
```
PatrickRudolph: just iterator over all tests in a single loop and then compare idx with testnos

flanfly:
Rebased and merged

PR #57, opened 2019-08-22
Split TPMPresent tests into three seperate tests.

ChriMarMe:
Made TestTPM12Present and TestTPM2Present optional
Third Test is required and checks that at least one connection is available

### DIFF
--- a/pkg/test/tpm.go
+++ b/pkg/test/tpm.go
@@ -26,10 +26,24 @@ var (
 		function: TestTPMConnect,
 		Status:   TestImplemented,
 	}
-	testtpmispresent = Test{
+	testtpm12present = Test{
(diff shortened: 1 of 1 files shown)


pkg/test/tpm.go
```diff
@@ -126,7 +126,7 @@ func TestTPM2Present() (bool, error) {
 }
 
 func TestTPMIsPresent() (bool, error) {
-	if (tpm12Connection != nil) || (tpm20Connection != nil) {
+	if (testtpm12present.Result.String() == "Pass") || (testtpm2present.Result.String() == "Pass") {
```
PatrickRudolph: == test.ResultPass

PR #61, opened 2019-08-23
Added some hopefully useful error msgs to tests.

ChriMarMe:


pkg/test/fit.go
```diff
@@ -237,8 +241,10 @@ func TestHasBIOSACM() (bool, error) {
 			count += 1
 		}
 	}
-
-	return count == 1, nil
+	if count == 0 {
```
PatrickRudolph: handle both count == 0 and count > 1

PatrickRudolph: count > 1 seems to be valid according to the spec.

PR #84, opened 2019-09-17
Add call to go fmt and go test to CI

ChriMarMe:
Run go fmt in every source code directory
Run go test in ./pkg/api/


ChriMarMe:
Testing hardware depended software configuration is a pain in the bubu

PR #85, opened 2019-09-18
Introduced modules as dependency manager

ChriMarMe:


.circleci/config.yml
```diff
@@ -16,8 +16,9 @@ jobs:
       - run: gofmt -e ./pkg/api/
       - run: gofmt -e ./pkg/test/
       - run: gofmt -e ./cmd/txt-suite/
-      - run: go test ./pkg/api/
       - run: go build -o txt-suite cmd/txt-suite/*.go
+      - run: go mod verify
```
PatrickRudolph: go mod verify needs to be run before build

.circleci/config.yml
```diff
@@ -16,8 +16,9 @@ jobs:
       - run: gofmt -e ./pkg/api/
       - run: gofmt -e ./pkg/test/
       - run: gofmt -e ./cmd/txt-suite/
```
PatrickRudolph: "go mod download" is missing

ChriMarMe: fixed that.

PR #95, opened 2020-06-23
Fix more TPM tests and implement IA32 debug interface check

zaolin:


pkg/test/cpu.go
```diff
@@ -292,3 +300,19 @@ func TestNoBIOSACMErrors() (bool, error, error) {
 
 	return !regs.ErrorCode.ValidInvalid, nil, nil
 }
+
+// IA32DebugInterfaceLockedDisabled checks if IA32 debug interface is locked
+func IA32DebugInterfaceLockedDisabled() (bool, error, error) {
```
ChriMarMe, line 305: This is a test, so put Test as prefix in function name

PR #96, opened 2020-06-24
API: Accept upper case "reserved" keyword

marcoguerri:
Depending on the kernel version, it might be uppercase or lowercase. 

See https://github.com/torvalds/linux/commit/640e1b38b00550990cecd809021cd37716e45922.

### DIFF
--- a/pkg/api/e820.go
+++ b/pkg/api/e820.go
@@ -5,9 +5,19 @@ import (
 	"io/ioutil"
 	"os"
 	"strconv"
+	"strings"
 )
 
-const reservedKeyword = "Reserved\n"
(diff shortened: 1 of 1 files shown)


marcoguerri:
Note: if I understand correctly,  `TestE820ReservedCheck` goes directly for the system's memmap, which is not mocked. So this is not really unit test.

To implement unit tests, I'd need to change a bit how `IsReservedInE810` reads the filesystem, to be able to do dependency injection. Let me know if you want me to give it a try.

zaolin:
@marcoguerri I am going to refactor it, Feel free to push PRs :)

PR #124, opened 2020-07-15
Fix return calls, now with proper error msgs if failing.

ChriMarMe:


pkg/test/cpu.go
```diff
@@ -294,8 +308,10 @@ func NoBIOSACMErrors(txtAPI hwapi.APIInterfaces) (bool, error, error) {
 	if err != nil {
 		return false, nil, err
 	}
-
-	return !regs.ErrorCode.ValidInvalid, nil, nil
+	if regs.ErrorCode.ValidInvalid {
+		return false, fmt.Errorf("BIOS ACM had start up errors"), nil
```
zaolin, line 312: Check for no error and return error at the end

ChriMarMe: done

Issue #142, opened 2020-07-30
Release 2.1


zaolin:
- [ ] ISO image #139 

- [x] TXT provisioning tooling

- [ ] Minor bug #140 #129 #123 #122 

- [x] Introduce ranges and selection together in the command line parser

- [x] Introduce a check for coreboot which deselects tests

- [x] Remove SINIT errors test

- [x] Improve test run output

- [x] Remove WARN level #137

PR #143, opened 2020-07-30
Fix tests and command line tool

zaolin:
See #142


pkg/test/memory.go
```diff
@@ -124,13 +124,6 @@ var (
 		SpecificiationTitle:     IntelTXTSpecificationTitle,
 		SpecificationDocumentID: IntelTXTSpecificationDocumentID,
 	}
-	testnosiniterrors = Test{
-		Name:        "SINIT ACM startup successful",
-		Required:    false,
-		NonCritical: true,
-		function:    NoSINITErrors,
-		Status:      Implemented,
-	}
```
ChriMarMe, line 133: Why was this removed?

pkg/test/test.go
```diff
@@ -54,15 +54,23 @@ const (
...
+const (
+	// FWCoreboot indicates test does run on coreboot firmware
+	FWCoreboot FirmwareType = "coreboot"
+
+	// FWNoCoreboot indicates test doesn't run on coreboot firmware
+	FWNoCoreboot FirmwareType = "UEFI"
+)
+
 func (t Result) String() string {
-	return [...]string{"TESTNOTRUN", "DEPENDENCY_FAILED", "INTERNAL_ERROR", "FAIL", "WARN", "PASS"}[t]
```
ChriMarMe, line 65: Why remove the concept of warn?

pkg/test/tpm.go
```diff
@@ -895,39 +846,26 @@ func NPWModeIsNotSetInPS(txtAPI hwapi.APIInterfaces) (bool, error, error) {
...
+			return true, fmt.Errorf("Auto-promotion mode active"), nil
 		}
 	}
 	if pol2 != nil {
+		if pol2.PolicyType != tools.LCPPolicyTypeAny {
+			return true, fmt.Errorf("Signed Policy mode active"), nil
+		}
 		if pol2.PolicyType != tools.LCPPolicyTypeList {
-			return false, fmt.Errorf("Auto-promotion mode active"), nil
+			return true, fmt.Errorf("Auto-promotion mode active"), nil
```
ChriMarMe, line 868: Why we dont keep the granularity?

pkg/test/memory.go
ChriMarMe: Also seperate commit

pkg/test/test.go
ChriMarMe: Also seperate commit.

pkg/test/tpm.go
ChriMarMe: Maybe you guessed it ;) Separate commit

pkg/test/cpu.go
```diff
@@ -287,7 +286,7 @@ func IBBIsTrusted(txtAPI hwapi.APIInterfaces) (bool, error, error) {
 	if regs.BootStatus&(1<<59) != 0 && regs.BootStatus&(1<<63) != 0 {
 		return true, nil, nil
 	}
-	return false, fmt.Errorf("IBB not trusted"), err
+	return true, fmt.Errorf("IBB not trusted, MLE not running"), err
```
PatrickRudolph, line 289: That's impossible. If something returns true all error must be nil.

pkg/test/acpi.go
```diff
@@ -19,6 +19,7 @@ var (
 		Required:                true,
 		function:                CheckRSDPValid,
 		Status:                  Implemented,
+		Firmware:                FWNoCoreboot,
```
PatrickRudolph, line 22: The new Field Firmware is just wrong, whatever it does. That's not clear from the commit message.

pkg/tools/smbios.go
```diff
@@ -0,0 +1,27 @@
+package tools
+
+import (
+	"errors"
+
+	"github.com/digitalocean/go-smbios/smbios"
+)
+
+// SMBIOSGetVendor gets the vendor name from table 0
+func SMBIOSGetVendor() (*string, error) {
```
PatrickRudolph, line 10: that must be astracted over the hwapi.

PR #149, opened 2020-08-07
Improve txt-suite command-line tool output

zaolin:
* More compact command-line output
* Adding test numbers
* Remove document information
* Remove useless newlines 

Signed-off-by: Philipp Deppenwiese <zaolin@das-labor.org>

### DIFF
--- a/cmd/txt-suite/main.go
+++ b/cmd/txt-suite/main.go
@@ -66,6 +66,7 @@ func run(tests []*test.Test) bool {
 		ioutil.WriteFile(logfile, data, 0664)
 	}
 
+	fmt.Printf("For more information about the documents and chapters, run: txt-suite -m\n\n")
 	for index := range tests {
 		if tests[index].Status == test.NotImplemented {
 			continue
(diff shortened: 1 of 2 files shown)


cmd/txt-suite/main.go
```diff
@@ -66,28 +66,29 @@ func run(tests []*test.Test) bool {
...
-			fmt.Printf("%-20s\n", a.Bold(a.Yellow(tests[index].Result)))
+			fmt.Printf("%-20s", a.Bold(a.Yellow(tests[index].Result)))
 		} else {
-			fmt.Printf("%-20s\n", a.Bold(a.Red(tests[index].Result)))
+			fmt.Printf("%-20s", a.Bold(a.Red(tests[index].Result)))
 		}
 		if tests[index].ErrorText != "" {
-			fmt.Printf(" %s\n", tests[index].ErrorText)
-		}
-		if tests[index].ErrorTextSpec != "" {
```
PatrickRudolph, line 89: ErrorTextSpec has been removed, but no reason was given why.

zaolin: Output looks too complex. Document ids can be looked up via testplan

PatrickRudolph: too complex? Also the testplan isn't part of the binary.
How about adding a new flag (-v for verbose) and only print if if -v is specified.

zaolin: We already have a -m option which prints all necessary information.

PatrickRudolph: ????????????

PR #154, opened 2020-09-03
Fix provisioning tooling

zaolin:


pkg/provisioning/auxdelete.go
```diff
@@ -8,14 +8,15 @@ import (
 )
 
 // DeleteAUXindexTPM20 deletes the AUX index on TPM 2.0
-func DeleteAUXindexTPM20(rw io.ReadWriter, psPol *tools.LCPPolicy2, writeHash, delHash []byte) error {
+func DeleteAUXindexTPM20(rw io.ReadWriter, psPol *tools.LCPPolicy2, passHash []byte) error {
 	if (1 >> (psPol.PolicyControl & tools.LCPPolicyControlAuxDelete)) != 0 {
```
walterchris, line 12: ?

walterchris (commented):
Inconsistent use of

> if err := ... ; err !=nil { .. and err := .. if err != nil

walterchris (approved):

PR #158, opened 2020-09-08
Fix 122: Test "TXT register space accessible" should not check if HeapBase is 0 

ChriMarMe:


PatrickRudolph (commented):
makes no sense at all

PR #183, opened 2021-02-21
Introduce ugly hack to read data from images with reducted structure.

ChriMarMe:
Problem is, that certain images with a reducted structure won't parse.
This "patch" solves the problem in the short term for functionality.


ChriMarMe:
I know, I know, i should not write code in autogen files, but as I said......the ugliest hack just to make it work

ChriMarMe:
Oh, obviously it's so whack it also breaks the CI. Fabulous

ChriMarMe:
Mistake.

PR #185, opened 2021-02-22
Fix upstreaming issues, get tooling in shape

zaolin:


pkg/intel/metadata/manifest/key/manifest_nocodegen.go
```diff
@@ -11,24 +11,5 @@ import (
...
-	if m.KeyAndSignature.Signature.DataTotalSize() < 1 {
-		fmt.Printf("  --Key and Signature--\n")
-		if m.KeyAndSignature.Key.DataTotalSize() > 0 {
-			fmt.Printf("\t%v\n", m.KeyAndSignature.Key.PrettyString(2, true))
-		}
-		fmt.Printf("\n\n\tKey Manifest not signed!\n\n")
-	} else {
-		fmt.Printf("%v \n\n", m.KeyAndSignature.PrettyString(2, true))
-	}
+	fmt.Printf("%v\n", m.PrettyString(1, true))
```
ChriMarMe, line 14: This shows empty structures and with this implies a bpm may not be signed. @walterchris wanted me to have explicit information shown.

pkg/intel/metadata/manifest/crypto_routines.go
```diff
@@ -0,0 +1,109 @@
...
+	case AlgSM2:
+		_, err = s.WriteString("SM2")
+	default:
+		return fmt.Sprintf("Alg?<%d>", int(a))
+	}
+	if err != nil {
+		return fmt.Sprintf("Writing to string builder failed: %v", err)
+	}
+	return s.String()
+}
```
ChriMarMe, line 109: This belongs in pkg/provisioning/bg

xaionaro: I remember already introducing Algorithms into `manifest`. But it was erased. May be it makes sense to just revert?

It is I'm just thinking out loud :)

pkg/intel/metadata/manifest/key/manifest_nocodegen.go
zaolin: Fixed by @xaionaro and me

pkg/intel/metadata/manifest/crypto_routines.go
zaolin: I did that because the go-tpm library doesn't implement the necessary algorithms and we just make a tiny use of the code. Also the CBnT crypto is somewhat special..

pkg/intel/metadata/manifest/key.go
```diff
@@ -131,19 +152,59 @@ func (k *Key) SetPubKey(key crypto.PublicKey) error {
...
+			if err := binary.Write(buf, binary.LittleEndian, k.Data[4:]); err != nil {
+				return err
+			}
+			if err := binary.Write(buf, binary.LittleEndian, k.Data[:4]); err != nil {
+				return err
+			}
+			if kmAlg != AlgSHA256 {
+				return fmt.Errorf("KM public key hash algorithm must be SHA256")
+			}
+			hash, err := kmAlg.Hash()
```
ChriMarMe, line 199: Is there an indicator in spec, that the hash function can be SHA384 or SM3?
If so:

```
if kmAlg == AlgSHA256 || kmAlg == AlgSHA384 || kmAlg == AlgSM3 {
    hash, err := kmAlg.Hash()
    if err != nil {
        return err 
    }
} else {
	return fmt.Errorf("KM public key hash algorithm must be SHA256")
}
hash.Write(buf.Bytes())
``

ChriMarMe (approved):

xaionaro (approved):

pkg/intel/metadata/manifest/key.go
walterchris: KM pubkey digest can be either SHA256 or SM3 to fit into 32 bytes of FPFs. SHA1
digest despite that it fits into 32 bytes is not supported since it was historically
forbidden in HSW and SKL platforms.
From the BIOS Specification #575623

PR #195, opened 2021-03-27
Tpm12 support

rihter007:
Add new FlowIntelLegacyTXTEnabledTPM12 flow
Try to detect TPM12 based on registers and firmware


rihter007:
I want to notice that build fails because of a different reason :)
I tested the code in a private repo

rihter007:
Will add BTG_SACM_INFO register, it also contains information about TPM Type and is more stable than ACM_POLICY_STATUS

pkg/pcr/flow_detection.go
```diff
@@ -57,7 +144,7 @@ func isCBnT(fitEntries []fit.Entry) (bool, error) {
 		case *fit.EntryKeyManifestRecord:
 			data, err := fitEntry.ParseData()
 			if data == nil {
-				return false, fmt.Errorf("unable to parse KeyManifest policy record: %w", err)
+				return false, fmt.Errorf("unable to parse KeyManifest policy record: %v", err)
```
xaionaro, line 147: why? :)

pkg/pcr/flow_detection.go
```diff
@@ -1,13 +1,96 @@
...
+				if s == 1 {
+					if tpmFamilySupport.IsDiscreteTPM12Supported() {
+						return registers.TPMType12, nil
+					}
+					if tpmFamilySupport.IsDiscreteTPM20Supported() {
+						return registers.TPMType20, nil
+					}
+					if tpmFamilySupport.IsFirmwareTPM20Supported() {
+						return registers.TPMTypeIntelPTT, nil
+					}
```
xaionaro, line 85: nitpicking:

I would use `switch` instead

pkg/pcr/flow_detection.go
```diff
@@ -1,13 +1,96 @@
...
+				// if none options is set - no TPM
+				// if only one option is set - can determine
+				s := bool2Int(tpmFamilySupport.IsDiscreteTPM20Supported()) +
+					bool2Int(tpmFamilySupport.IsFirmwareTPM20Supported()) +
+					bool2Int(tpmFamilySupport.IsDiscreteTPM12Supported())
+
+				if s == 0 {
+					return registers.TPMTypeNoTpm, nil
+				}
+				if s == 1 {
```
xaionaro, line 76: What should we do if `s > 1`?

xaionaro (approved):
Looks good!

pkg/pcr/flow_detection.go
rihter007: I thought that it is a typo, checked the docs, %w is now used for wrapping, will revert this line, thanks!

pkg/pcr/flow_detection.go
rihter007: Suppose tpm2 by default...

I also have a PR where we get user-defined TPM value. Only if it is not set, this fallback detection will be used.
So, the PCR0 calc algorithm will take the following tuple: <required firmware, optional registers, optional tpm_version>.
If something is not provided, just try our best (like this fallback) with the data we have :)

PR #197, opened 2021-03-27
Simple heuristic for local TPM version detection

rihter007:
I noticed that the code I need is located here: https://github.com/9elements/tpmtool/blob/master/pkg/tpm/tpm.go
But it will not work for me for several reasons:
- I do not need to open TPM device + opening TPM device may fail (due to resource broker)
- /sys/class/tpm/tpm0/caps file does not exist for TPM2.0, but the code path will fail if this is the case

So I decided to create a simple package here in order not to get blocked.


xaionaro (changes requested):
Too huge change.

ChriMarMe:
Rebase on master

rihter007:
> Too huge change.

The change is small (50 lines of code), just pcr0tool was reverted

PR #199, opened 2021-03-29
Integration of PCR0Tool 

ChriMarMe:


rihter007:
Hm, it seems that the build fails because the dependency "github.com/dustin/go-humanize" is missing from go.mod

ChriMarMe:
github.com/dustin/go-humanize v1.0.0 exists in the go.mod.

rihter007:
I also do not see changes in pkg/hwapi (I had some for registers pkg)

b6e101712b413aac677ce95d0dd789fbf31fe380
3d8a557b77c8308c06fb37297edb0268bb289b1f
9c6c12fc1b879833086c63a866942eec6b3e2707
1b42bd9b3b8f635f79db05a48c5953afc05b9f1c

rihter007:
> github.com/dustin/go-humanize v1.0.0 exists in the go.mod.

Hm, you are right. It is weird that the build is failing because of it

ChriMarMe:
I guess the problem is that there are still some "v2" mentions in import pathes. but these "v2" is only required for versioning, not importing.

xaionaro:
> It is weird that the build is failing because of it

The codegen I wrote does not work with Go modules. It looks into the main go/src tree instead :(
So it is required to add `GO111MODULE=off go get github.com/dustin/go-humanize` to make the CI-test work.

ChriMarMe:
If I remember correctly, you fixed that in the past. Why does it break now?

ChriMarMe:
Problem caused by file

> tpm_info_list_manifestcodegen.go

There is a issue with the import of 

> /go/src/github.com/9elements/converged-security-suite/v2/pkg/intel/metadata/manifest/common/pretty

I can't figure out what causes the problem.

xaionaro:
I'll take a look.

ChriMarMe:
> I guess the problem is that there are still some "v2" mentions in import pathes. but these "v2" is only required for versioning, not importing. 

This is definitly working as intendet. Sorry for wrong assumption here. "v2" in the path is required.

xaionaro (approved):
Great job!

rihter007 (approved):
Excellent!

xaionaro (approved):
Re-approving

ChriMarMe:
The code also touches parts which have other code owners. @zaolin @walterchris needs to review this as well.

walterchris (approved):

zaolin (approved):

PR #207, opened 2021-03-30
[breaking] Rewrite bg-prov stitch

rojer9-fb:
* Move args to optional flag for better extensibility
 * Add ability to replace ME region

What used to be

```
bg-prov coreboot.rom acm.bin km.bin bpm.bin
```

becomes

```
bg-prov coreboot.rom --acm=acm.bin --km=km.bin --bpm=bpm.bin
```

Additionally, it is now possible to replace ME with `--me=me.bin`

Verified that `--me=x` produces binary identical to `ifdtool -i ME:x`

### DIFF
--- a/cmd/bg-prov/cmd.go
+++ b/cmd/bg-prov/cmd.go
@@ -11,6 +11,8 @@ import (
 	"io/ioutil"
 	"os"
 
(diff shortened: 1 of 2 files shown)


walterchris (dismissed):
LGTM

walterchris (approved):

PR #221, opened 2021-04-16
Suppress fiano warning

xaionaro:
Fiano prints logs and this is non-disablable. See https://github.com/linuxboot/fiano/issues/330

In this PR I temporary suppress fiano-related logs when we call `tools.GetRegion`.

`log.SetOutput` uses a mutex, so it should be safe:
```
// SetOutput sets the output destination for the standard logger.
func SetOutput(w io.Writer) {
	std.mu.Lock()
	defer std.mu.Unlock()
	std.out = w
}
```

### DIFF
--- a/pkg/tools/fiano_suppress_log.go
+++ b/pkg/tools/fiano_suppress_log.go
@@ -1,0 +1,30 @@
+package tools
+
+import (
+	"io"
+	"log"
+	"runtime"
+	"strings"
+)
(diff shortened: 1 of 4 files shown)


ChriMarMe (approved):

xaionaro:
A better fix: https://github.com/linuxboot/fiano/pull/331

PR #228, opened 2021-04-28
Fix MergeRanges functions fo the case when a longer range with a shorter one

rihter007:


pkg/bytes/range.go
```diff
@@ -13,7 +13,7 @@ type Range struct {
 }
 
 func (r Range) String() string {
-	return fmt.Sprintf(`{"Offset":"0x%x", "Length":"0x%x"}`, r.Offset, r.Length)
+	return fmt.Sprintf(`{"Offset":"%d", "Length":"%d"}`, r.Offset, r.Length)
```
xaionaro, line 16: If this is integers, they might be without quotes.

xaionaro (dismissed):

pkg/bytes/range.go
rihter007: Sorry, I did it for local debug

ChriMarMe (approved):

xaionaro (approved):

PR #232, opened 2021-05-05
CBnT/feature/cbfs support 2

ArthurHeymans:


ChriMarMe (dismissed):

cmd/cbnt-prov/cmd.go
```diff
@@ -416,11 +416,13 @@ func (g *generateBPMCmd) Run(ctx *context) error {
 			se.DigestList.List[iterator].HashAlg = g.IbbHash[iterator]
 		}
 
-		seg := *bootpolicy.NewIBBSegment()
-		seg.Base = g.IbbSegbase
-		seg.Size = g.IbbSegsize
-		seg.Flags = g.IbbSegFlag
-		se.IBBSegments = append(se.IBBSegments, seg)
+		if g.IbbSegbase != 0 {
```
ChriMarMe, line 419: I think this will break if you don't supply -iibsegbase=0

pkg/provisioning/cbnt/tools.go
```diff
@@ -419,3 +421,42 @@ func StitchFITEntries(biosFilename string, acm, bpm, km []byte) error {
...
+// additional IBBSegment.
+func FindAdditionalIBBs(imagepath string) ([]bootpolicy.IBBSegment, error) {
+	ibbs := make([]bootpolicy.IBBSegment, 0)
+	image, err := os.Open(imagepath)
+	if err != nil {
+		return nil, err
+	}
+	defer image.Close()
+
+	imageI, err := image.Stat()
```
walterchris, line 435: ImageI ?

cmd/cbnt-prov/cmd.go
ChriMarMe: I had the same issue several times in the past. If you check on the existence of an optional argument which wasn't supplied, it is not set at all and will result in a null pointer exception.

pkg/provisioning/cbnt/tools.go
ArthurHeymans: imageI(nformation). I'll find a better name

cmd/cbnt-prov/cmd.go
ArthurHeymans: This seems to work fine, as coreboot does not specify all arguments. kong does have a flags to set the default value of a command line argument. Should that be used?

ChriMarMe: If it works it works

ChriMarMe (approved):

cmd/cbnt-prov/cmd.go
ArthurHeymans: Are variables not 0 initialized in go? That's probably why it works then for integers.

pkg/provisioning/cbnt/tools.go
ArthurHeymans: Done

ChriMarMe (approved):

walterchris (approved):

PR #234, opened 2021-05-07
fix(prov): Allow export of ACM even when KM/BPM unavailable

xaionaro:
We need to export an ACM on a pre-CBnT machine with no BPM/KM. But also if there is a CBnT machine with damaged FIT (such that BPM is not available) it makes sense to permit exporting ACM.

# Test plan

Before:
```
[root@twdi46715.ftw2 ~]# ./cbnt-prov acm-export current.fd current.acm
bg-prov: error: image has no BPM (isNil:true) or/and KM (isNil:true) or/and ACM (isNil:false)
```

After:
```
[root@twdi46715.ftw2 /tmp]# ./cbnt-prov acm-export ~/current.fd acm
[root@twdi46715.ftw2 /tmp]# ls -ld acm
-rw-r--r-- 1 root root 131072 May  7 05:26 acm
```


ChriMarMe (approved):

pkg/provisioning/cbnt/tools.go
```diff
@@ -125,7 +125,7 @@ func ParseFITEntries(image []byte) (bpm *fit.EntryBootPolicyManifestRecord, km *
 		}
 	}
 	if bpm == nil || km == nil || acm == nil {
-		return nil, nil, nil, fmt.Errorf("image has no BPM (isNil:%v) or/and KM (isNil:%v) or/and ACM (isNil:%v)", bpm == nil, km == nil, acm == nil)
```
rihter007, line 128: Nit:
It is not common to get an error and valid data...
I would rather remove error as a part of function's result completely (in this case we miss possible handling of the error) or do not return error on partial results (remove lines 127-129)

rihter007 (approved):
Minor nit

pkg/provisioning/cbnt/tools.go
xaionaro: I agree. I had the same thoughts, but I did want to focus on the main task instead of polishing the code here.

PR #235, opened 2021-05-11
testdata: Add test keys

ArthurHeymans:
Those can be used in unit tests and for the coreboot buildsystem
integration.

Signed-off-by: Arthur Heymans <arthur@aheymans.xyz>


ChriMarMe (dismissed):

ArthurHeymans:
> Instead of adding .pem key - shouldn't we add priv and pub key separately?

I used pem keys for both BtG and CBnT and never tried separate priv and pub keys. Does the tooling work with that?

ChriMarMe:
Yes, the tooling works with priv and pub keys in pem format in seperate files

ArthurHeymans:
> Instead of adding .pem key - shouldn't we add priv and pub key separately?

Done

ChriMarMe (approved):

walterchris (approved):

PR #237, opened 2021-05-11
tools/ifd: Fall back to parsing coreboot FMAP when no IFD is found

ArthurHeymans:
This makes it possible to operate on a coreboot BIOS-region-only
image. This is useful to allow buildtesting coreboot without an IFD.

Signed-off-by: Arthur Heymans <arthur@aheymans.xyz>


pkg/tools/ifd.go
```diff
@@ -1,18 +1,41 @@
...
+	if ifdErr == nil {
+		return uint64(off+size) - consts.BasePhysAddr + addr, nil
+	}
+	// If no IFD is present we are not dealing with a full image,
+	// but maybe a BIOS region only image. Attempt to parse coreboot fmap.
+	off, size, cbErr = getCorebootRegion(image)
+	// If it's not a proper coreboot image return the IFD error as this is more generic.
+	if cbErr != nil {
+		return 0, ifdErr
 	}
```
ChriMarMe, line 39: At the beginning of the function:
```
var cbErr error
```

PR #239, opened 2021-05-17
fix(pcr): Validate IBB signature in CBnT-0T

xaionaro:


cmd/pcr0tool/commands/printnodes/command.go
```diff
@@ -82,6 +82,7 @@ func (cmd Command) Execute(args []string) {
 	}
 	imagePath := args[0]
 
+	fianoUEFI.DisableDecompression = false
```
xaionaro, line 85: It appears `printnodes` wasn't printing compressed nodes since the `DisableCompression` optimization was merged. Fixing it.

pkg/intel/metadata/manifest/crypto_routines.go
```diff
@@ -41,14 +40,14 @@ const (
...
-	{AlgSHA256, crypto.SHA256.New()},
-	{AlgSHA384, crypto.SHA384.New()},
-	{AlgSHA512, crypto.SHA512.New()},
-	{AlgSM3_256, sm3.New()},
+	{AlgSHA1, crypto.SHA1.New},
+	{AlgSHA256, crypto.SHA256.New},
+	{AlgSHA384, crypto.SHA384.New},
+	{AlgSHA512, crypto.SHA512.New},
+	{AlgSM3_256, sm3.New},
 }
```
xaionaro, line 51: There was a bug: if I call `Algorithm.Hash()` then I get a hash function which sometimes is already in a dirty state. Fixing.

pkg/pcr/measurement.go
```diff
@@ -48,6 +48,19 @@ func (chunk DataChunk) String() string {
...
+	for idx := range s {
+		d := &s[idx]
+		if d.ID == id {
+			return d
+		}
+	}
+
+	return nil
+}
+
```
xaionaro, line 63: Was required while testing if IBB is indeed verified correctly (in our internal code).

pkg/pcr/validate_manifests.go
```diff
@@ -39,5 +39,9 @@ func (v ValidateManifests) Validate(firmware Firmware) error {
 		return fmt.Errorf("key chain is invalid: %w", err)
 	}
 
+	if err := bpm.ValidateIBBs(firmware); err != nil {
+		return fmt.Errorf("IBB signature in BPM is not valid: %w", err)
+	}
+
```
xaionaro, line 45: The fix itself.

pkg/registers/txt_test.go
```diff
@@ -109,7 +109,7 @@ func TestTXTErrorStatus(t *testing.T) {
 	if txtErrorStatus.BitSize() != 8 {
 		t.Errorf("Incorrect BitSize, expected: 64. actual: %v", txtErrorStatus.BitSize())
 	}
-	if txtErrorStatus.Address() != registers.TxtPublicSpace+registers.TXTStatusRegisterOffset {
+	if txtErrorStatus.Address() != registers.TxtPublicSpace+registers.TXTErrorStatusRegisterOffset {
```
xaionaro, line 112: Fixing `go test ./...`

ChriMarMe (approved):

ChriMarMe (approved):

zaolin (approved):

pkg/intel/metadata/manifest/bootpolicy/manifest.go
```diff
@@ -27,3 +32,28 @@ type Manifest struct {
 func (bpm Manifest) StructInfo() StructInfo {
 	return bpm.BPMH.StructInfo
 }
+
+func (bpm *Manifest) ValidateIBBs(firmware uefi.Firmware) error {
+	if len(bpm.SE[0].DigestList.List) == 0 {
```
rihter007, line 37: I would also put a check that len(bpm.SE) > 0
+ why are we doing only for the first element?

rihter007 (approved):

cmd/pcr0tool/commands/printnodes/command.go
rihter007: Could you please add this as a coment?

pkg/intel/metadata/manifest/bootpolicy/manifest.go
xaionaro: Unfortunately, the explanation is covered by NDA enforced by one of CPU vendors.

xaionaro: > I would also put a check that len(bpm.SE) > 0

To do not reset approvals for this tiny fix, I'll merge as is :(

PR #240, opened 2021-05-18
New PrintFit in tabular form

ChriMarMe:
Remove old PrintFIT function
Add new PrintFit function which prints FIT in a tabular form.

### DIFF
--- a/cmd/cbnt-prov/cmd.go
+++ b/cmd/cbnt-prov/cmd.go
@@ -11,6 +11,8 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/9elements/converged-security-suite/v2/pkg/intel/metadata/fit"
+
 	"github.com/linuxboot/fiano/pkg/uefi"
 
 	"github.com/9elements/converged-security-suite/v2/pkg/intel/metadata/manifest"
@@ -189,6 +191,10 @@ type keygenCmd struct {
 	Path     string `flag optional name:"path" help:"Path to store keys. File names are 'yourname_bpm/yourname_bpm.pub' and 'yourname_km/yourname_km.pub' respectivly"`
 }
 
+type printFITCmd struct {
+	BIOS string `arg required name:"bios" help:"Path to the full BIOS binary file." type:"path"`
+}
+
 func (v *versionCmd) Run(ctx *context) error {
 	tools.ShowVersion(programName, gittag, gitcommit)
 	return nil
@@ -256,10 +262,11 @@ func (biosp *biosPrintCmd) Run(ctx *context) error {
 	if err != nil {
 		return err
 	}
-	err = cbnt.PrintFIT(data)
+	table, err := fit.GetTable(data)
 	if err != nil {
 		return err
 	}
+	fmt.Printf("%s", table.String())
 	err = cbnt.PrintCBnTStructures(data)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


pkg/tools/fit.go
```diff
@@ -0,0 +1,25 @@
+package tools
+
+import (
+	"fmt"
+
+	"github.com/9elements/converged-security-suite/v2/pkg/intel/metadata/fit"
+)
+
+// PrintFit prints the Firmware Interface Table in a tabular human readable form.
+func PrintFit(image []byte) error {
```
xaionaro, line 10: According to https://github.com/golang/go/wiki/CodeReviewComments#initialisms it should be `PrintFIT` :)

xaionaro (dismissed):

pkg/tools/fit.go
```diff
@@ -0,0 +1,25 @@
...
+	fmt.Println("------------------------")
+	fmt.Printf("%-25s | %-20s | %-8s | %-10s | %-15s | %-10s\n", "Type", "Address", "Size", "Version", "Checksum valid", "Checksum")
+	fmt.Printf("-----------------------------------------------------------------------------------------------------\n")
+	for _, entry := range fit {
+		h := entry.GetHeaders()
+
+		fmt.Printf("%-25s | %-20s | %-8d | %-10s | %-15v | %-10d\n", h.Type().String(), h.Address.String(), h.Size.Uint32(), h.Version.String(), h.IsChecksumValid(), h.Checksum)
+	}
+	return nil
+}
```
xaionaro, line 25: IMHO, it would be better to place this function to `pkg/intel/metadata/fit` (instead of `pkg/tools`). It could be the `String` method of `fit.Table`. Also there's no need to call `GetEntries`: I see you print only headers, so it is enough to call `GetTable` (it will get only headers and won't parse referenced data).

pkg/tools/fit.go
ChriMarMe: Thanks for the review :) fixed it

ChriMarMe:
How about now?

xaionaro (dismissed):

pkg/intel/metadata/fit/table.go
```diff
@@ -22,6 +23,31 @@ func (table Table) GetEntries(firmware []byte) (result []Entry) {
 	return
 }
 
+// Print prints the fit table in a tabular form
```
xaionaro, line 26: A typo: Print -> String

pkg/intel/metadata/fit/table.go
```diff
@@ -22,6 +23,31 @@ func (table Table) GetEntries(firmware []byte) (result []Entry) {
...
+func (table Table) String() string {
+	var s strings.Builder
+	// PrintFit prints the Firmware Interface Table in a tabular human readable form.
+	s.WriteString("Firmware Interface Table\n")
+	s.WriteString("------------------------\n")
+	s.WriteString(fmt.Sprintf("%-25s | ", "Type") +
```
xaionaro, line 32: Just in case: it is also possible to do `fmt.Fprintrf(&s, ...)` instead of `s.WriteString(fmt.Sprintf(...))` if you would like this way.

xaionaro (dismissed):

xaionaro:
> How about now?

Looks nice. Thanks :)

pkg/intel/metadata/fit/table.go
ChriMarMe: I will try that

ChriMarMe: Done

cmd/cbnt-prov/cmd.go
```diff
@@ -805,6 +825,8 @@ var cli struct {
 	ACMExport acmExportCmd `cmd help:"Exports ACM structures from BIOS image into file"`
 	ACMShow   acmPrintCmd  `cmd help:"Prints ACM binary in human-readable format"`
 
+	FITShow printFITCmd `cmd help:"Prints the FIT Table of given BIOS image file"`
```
walterchris, line 828: nit: lowercase t? :)

xaionaro (approved):

walterchris (approved):

PR #246, opened 2021-06-10
feature(pcr): Add fake measurement for ACM

xaionaro:
The problem is that a bitflip in ACM causes TXT to disable. Therefore we should at least display these kind of problems to end users (for them be able to understand why TXT is disabled).

So I add the whole "ACM" as a fake measurement (which does not affect PCR0, but being displayed in reports).


dimkup (approved):
LGTM

rihter007 (approved):
LGTM

zaolin (approved):

PR #251, opened 2021-06-23
bugfix(registers): Avoid reading of ACM_POLICY_STATUS corruptor-register

xaionaro:
The problem is when we read this register (putted to the deny-list), it corrupts ACM_POLICY_STATUS (an undocumented behavior from PCH's side). Trying to avoid reading this register.


rihter007:
I would propose just to implement this hack here:

//FetchTXTConfigSpace returns a raw copy of the TXT config space
func FetchTXTConfigSpace(txtAPI PhysicalMemoryReader) (TXTConfigSpace, error) {
	data := make([]byte, TxtPublicSpaceSize)
	if err := txtAPI.ReadPhysBuf(TxtPublicSpace, data); err != nil {
		return nil, err
	}
	return data, nil
}

It might be even better to do:
// FetchTXTConfigSpace avoids memory corruption problem caused by...
func FetchTXTConfigSpaceSafe() ...

It will be closer to the problem, TxTAPI should be stay unmodified

ChriMarMe (approved):

pkg/registers/txt.go
```diff
@@ -29,15 +30,57 @@ type PhysicalMemoryReader interface {
...
+			// corruption.
+			Offset: 0xFED30370,
+			Length: 8,
+		},
+	}
+)
+
+// FetchTXTConfigSpaceSafe returns a filtered raw copy of the TXT config space,
+// it excludes registers, which is not supposed to be read (in contrast
+// to FetchTXTConfigSpaceSafe).
```
rihter007, line 64: nit: in contrast to FetchTXTConfigSpaceRaw

rihter007 (approved):

pkg/registers/txt.go
xaionaro: Cannot change without resetting the approvals :(

xaionaro: Will make an additional PR.

xaionaro: https://github.com/9elements/converged-security-suite/pull/253/files

PR #256, opened 2021-06-28
bugfix(pcr): Use only the first IBB digest for integrity validation

xaionaro:


rihter007 (approved):
LGTM, just add a comment why only the first segment is checked (method comment says that any of IBB segments doesn't match its signature)

ChriMarMe (approved):

PR #259, opened 2021-06-29
Make PrintRegisters/PrintRegister exportable

rihter007:
We can reuse these functions

### DIFF
--- a/cmd/pcr0tool/commands/dumpregisters/command.go
+++ b/cmd/pcr0tool/commands/dumpregisters/command.go
@@ -10,12 +10,6 @@ import (
 	"github.com/9elements/converged-security-suite/v2/pkg/registers"
 )
 
-func assertNoError(err error) {
-	if err != nil {
-		panic(err)
-	}
-}
-
 // Command is the implementation of `commands.Command`.
 type Command struct {
 	outputFile *string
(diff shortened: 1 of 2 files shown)


cmd/pcr0tool/commands/dumpregisters/command.go
```diff
@@ -64,7 +55,16 @@ func (cmd Command) Execute(args []string) {
...
+// PrintRegisters outputs registers in a detailed human-readable format
+func PrintRegisters(regs registers.Registers) {
+	for _, reg := range regs {
+		fmt.Printf("\n")
+		PrintRegister(reg)
+	}
+}
+
+// PrintRegister outputs a single register in a detailed human-readable format
+func PrintRegister(reg registers.Register) {
```
xaionaro, line 67: If we want to publish them, then it is required to move it to [`cmd/pcr0tool/commands/dumpregisters/helpers`](https://github.com/9elements/converged-security-suite/tree/master/cmd/pcr0tool/commands/dumpregisters/helpers). Otherwise there's too much dependencies for too lightweight logic.

xaionaro (changes requested):

ChriMarMe (approved):

xaionaro (approved):

PR #264, opened 2021-07-16
Add missing close to MSR register open fd

mimir-d:
- there are usecases where this missing close leads to fd exhaustion

testing:
```
% go test   
PASS
ok  	github.com/9elements/converged-security-suite/v2/pkg/registers	0.289sw
```

### DIFF
--- a/pkg/hwapi/msr.go
+++ b/pkg/hwapi/msr.go
@@ -30,14 +30,20 @@ type IA32Debug struct {
 	PCHStrap bool
 }
 
+func readMSRFromCpu(msr int64, cpu int) (uint64, error) {
+	msrCtx, err := gomsr.MSR(cpu)
+	if err != nil {
+		return 0, fmt.Errorf("MSR: Selected core %d doesn't exist (%w)", cpu, err)
+	}
+	defer msrCtx.Close()
+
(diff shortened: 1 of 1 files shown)


xaionaro (changes requested):
The linter is right, in this case we theoretically can get a panic.

Also may be it makes sense to put the whole content of the `for` to a separate function to avoid `defer`-ing in a loop. But this is a nitpicking. Feel free to ignore :)

rihter007:
> The linter is right, in this case we theoretically can get a panic.
> 
> Also may be it makes sense to put the whole content of the `for` to a separate function to avoid `defer`-ing in a loop. But this is a nitpicking. Feel free to ignore :)


I guess once can use inplace function for it:

for i := 0; i < runtime.NumCPU(); i++ {
    func() {
       ...
       defer msr.Close()
       ...
    }()
}

mimir-d:
i could've sworn i put that defer after the error check. fixing...

xaionaro (dismissed):
Thanks!

rihter007 (dismissed):
LGTM, only one optional nit

xaionaro (approved):

zaolin (approved):

PR #272, opened 2021-07-27
Remove acpi functions as members of interface structure. 

ChriMarMe:
They were never part of the interface


flanfly (approved):

xaionaro (approved):

flanfly:
Some of functions in the hwapi package use TxtAPI instead of APIInterfaces. Also the UintN structure isn't included in non x86 builds (phys.go has build flags). This patch split phys.go into phys_{le,be}.go and moves UintN into it's own file. The phys* split duplicates some code.

PR #281, opened 2021-08-12
feat(pcr): Report about corruptions in KM, BPM and IBB

xaionaro:
In CBnT there's no direct measurement of IBB, so if corruption happens in IBB we see signature validation error, but do not see what exactly causes it. Fixing this.

# Before

```
xaionaro@void:~/go/src/github.com/9elements/converged-security-suite$ go run ./cmd/pcr0tool/ diff ~/firmware/F0E_3A10.bin ~/firmware/F0E_3A10.bin-c.bin 2>/dev/null | grep -A 1000 Total
Total:
	changed bytes: 0 (in 0 ranges)
	hamming distance: 0
	hamming distance for non-(0x00|0xff) bytes: 0
The earliest offset of a different measured bytes: 0xffffffffffffffff
```

# After

```
xaionaro@void:~/go/src/github.com/9elements/converged-security-suite$ go run ./cmd/pcr0tool/ diff ~/firmware/F0E_3A10.bin ~/firmware/F0E_3A10.bin-c.bin 2>/dev/null | grep -A 1000 Total
Total:
	changed bytes: 1 (in 1 ranges)
	hamming distance: 8
	hamming distance for non-(0x00|0xff) bytes: 0
The earliest offset of a different measured bytes: 0x3e4cc67
```


ChriMarMe (approved):

pkg/intel/metadata/manifest/bootpolicy/manifest.go
```diff
@@ -78,3 +75,18 @@ func (bpm *Manifest) ValidateIBBs(firmware uefi.Firmware) error {
 
 	return nil
 }
+
+// IBBDataRanges returns data ranges of IBB.
+func (bpm *Manifest) IBBDataRanges(firmwareSize uint64) pkgbytes.Ranges {
+	var result pkgbytes.Ranges
+
+	for _, seg := range bpm.SE[0].IBBSegments {
+		if seg.Flags&1 == 1 {
```
rihter007, line 84: nit: it would be nice to enum this bit

rihter007 (approved):

PR #293, opened 2021-09-21
MeasureFunc returns a slice of measurements to support repeating measurements.

rihter007:
Interface change, the functionality should not be affected
Tested internally that behaviour of the measurements didn't change


ChriMarMe (approved):

xaionaro:
@rihter007 : I'm getting hard time to understand why :(

What is "repeating measurements"?

rihter007:
> @rihter007 : I'm getting hard time to understand why :(
> 
> What is "repeating measurements"?

Sorry for confusing. I was solving a task of measuring all of specific items in firmware. You can have a look on PR 295 for an example.

In general, firmware has undefined (could be 0 or 10) number of specific items. I want to measure all of them sequentially.

xaionaro:
> Decided to choose a better approach

Well, basically the approach is the same. Just with attempt to leave a hint that Measurements should be single :)

PR #294, opened 2021-09-22
Add Flow.{Platform,CPU}VendorID

xaionaro:
I need to add vendor-specific logic in the internal service, which uses pcr0tool. So I think this is a good place to add the mapping of flows to vendor IDs.


ChriMarMe:
Hmmmm. This could be part of github.com/9elements/go-linux-lowlevel-hw instead?

I think it's a more fitting place because this lib is taking care of hardware abstraction.

xaionaro:
You mean specifically package `platformid`?

ChriMarMe:
Yes

xaionaro:
Decided not to push this logic to CSS, yet. Need to polish up the entities internally first (to avoid mess in the upstream.)

PR #299, opened 2021-10-05
Fix AMD firmware parsing slice bounds out of range

orangecms:
The offset must be within the size of the image.


zaolin (approved):

pkg/amd/manifest/firmware.go
```diff
@@ -107,7 +107,7 @@ func parsePSPFirmware(firmware Firmware) (*PSPFirmware, error) {
 		efs.BIOSDirectoryTableFamily17hModels60h3FhPointer,
 	}
 	for _, offset := range biosDirectoryOffsets {
-		if offset == 0 {
+		if offset == 0 || int(offset) > len(image) {
```
rihter007, line 110: Good catch!

rihter007 (approved):

PR #300, opened 2021-10-07
AMD manifest: return meaningful errors for structures not found

orangecms:
The previous errors were the same for all cases and misleading.


orangecms:
closing this one, since this now lives in Fiano and came via https://github.com/linuxboot/fiano/pull/340

PR #308, opened 2021-12-01
Change MeasurementIDDXE to EV_EFI_PLATFORM_FIRMWARE_BLOB2

trynity:
On AMD Platforms, this particular measurement can be EV_EFI_PLATFORM_FIRMWARE_BLOB2
with the newer TCG 1.5 spec.

Ideally, we should be able to tell which version of the TCG spec is being
used for the firmware, but this can be done right now in order to
unblock.

### DIFF
--- a/pkg/pcr/measurement_id.go
+++ b/pkg/pcr/measurement_id.go
@@ -263,46 +263,48 @@ func (id MeasurementID) PCRIDs() []ID {
 	return nil
 }
 
-// EventLogEventType returns value of "Type" field of the EventLog entry
+// EventLogEventTypes returns multiple potential values of "Type" field of the EventLog entry
 // associated with the measurement.
-func (id MeasurementID) EventLogEventType() *tpmeventlog.EventType {
+func (id MeasurementID) EventLogEventTypes() []*tpmeventlog.EventType {
+	var eventTypes []*tpmeventlog.EventType
 	switch id {
 	case MeasurementIDInit:
-		return eventTypePtr(tpmeventlog.EV_NO_ACTION)
+		eventTypes = append(eventTypes, eventTypePtr(tpmeventlog.EV_NO_ACTION))
 	case MeasurementIDPCR0DATA:
-		return eventTypePtr(tpmeventlog.EV_S_CRTM_CONTENTS)
+		eventTypes = append(eventTypes, eventTypePtr(tpmeventlog.EV_S_CRTM_CONTENTS))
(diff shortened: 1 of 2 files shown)


trynity:
Going to add the ability to have multiple event types for the same measurement per @rihter007's feedback

xaionaro:
@rihter007, @trynity  : Just a thought: May be we should use different MeasurementID for this case?

rihter007:
> @rihter007, @trynity : Just a thought: May be we should use different MeasurementID for this case?

I remembered we had a discussion about it back in August. The reason behind supporting both of them is that both values fit the docs as @trynity mentioned.
So, as we desire to support generic behaviour, we should make this measurementID be aligned with both options

trynity:
@xaionaro @rihter007 Adjusted the approach now to allow multiple event types for a measurement, let me know if it should be further refined, I'm not quite sure how to do it in a more intelligent way (Such as detect the TCG spec version somehow)

pkg/pcr/measurement_id.go
```diff
@@ -322,16 +324,17 @@ func TPMEventTypeToMeasurementIDs(pcrID ID, tpmEventType tpmeventlog.EventType)
...
 		}
 
-		eventType := measurementID.EventLogEventType()
-		if eventType == nil {
-			continue
-		}
+		eventTypes := measurementID.EventLogEventTypes()
 
-		if *eventType != tpmEventType {
+		if len(eventTypes) == 0 {
```
rihter007, line 329: Nit, I would remove this check

pkg/pcr/measurement_id.go
```diff
@@ -263,46 +263,48 @@ func (id MeasurementID) PCRIDs() []ID {
 	return nil
 }
 
-// EventLogEventType returns value of "Type" field of the EventLog entry
+// EventLogEventTypes returns multiple potential values of "Type" field of the EventLog entry
 // associated with the measurement.
-func (id MeasurementID) EventLogEventType() *tpmeventlog.EventType {
+func (id MeasurementID) EventLogEventTypes() []*tpmeventlog.EventType {
```
rihter007, line 268: Nit: I would make return type: "[]tpmeventlog.EventType"

rihter007 (approved):
Just a couple of optional nits
GTG!

zaolin (approved):

trynity:
Will address the nits in a follow-up PR

PR #323, opened 2022-03-02
add DataChunkIDBIOSStartup4 and DataChunkIDBIOSStartup5

trynity:
Right now, we're just adding two additional startup modes
Longer term this would be better to be dynamic, since
according to the specification, there is no limit to the
number of Startup Modules.

From Intel's Documentation:

In order to enable more flexible flash layout, BIOS Init code can be split in multiple BIOS Startup Modules. Each BIOS Startup Module will have one corresponding Type 7 entry. Each Type 7 entry describes address and size of the corresponding BIOS Startup Module.

### DIFF
--- a/pkg/pcr/measurement_id.go
+++ b/pkg/pcr/measurement_id.go
@@ -705,6 +705,8 @@ const (
 	DataChunkIDBIOSStartup1
 	DataChunkIDBIOSStartup2
 	DataChunkIDBIOSStartup3
+	DataChunkIDBIOSStartup4
+	DataChunkIDBIOSStartup5
 	DataChunkIDACMPolicyStatus
 	DataChunkIDACMHeaderSVN
 	DataChunkIDACMSignature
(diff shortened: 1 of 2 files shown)


rihter007 (approved):
LGTM

pkg/pcr/measurement_id.go
```diff
@@ -716,18 +718,22 @@ const (
...
 	case 3:
-		return DataChunkIDBIOSStartup3
+		return DataChunkIDBIOSStartup3, nil
+	case 4:
+		return DataChunkIDBIOSStartup4, nil
+	case 5:
+		return DataChunkIDBIOSStartup5, nil
 	}
-	panic(fmt.Sprintf("should not happen: %v", entryIndex))
+	return DataChunkIDUndefined, fmt.Errorf("invalid DataChunkIDBIOSStartup: %v", entryIndex)
```
trynity, line 736: @rihter007 Returning an error now.

ChriMarMe (approved):

PR #325, opened 2022-04-13
[DRAFT][WIP] Working on adding support for Lenovo X1 Gen9

xaionaro:


cmd/pcr0tool/commands/displayeventlog/command.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if *cmd.calcPCR {
+		calculatedValue, err := pcr.Replay(eventLog, pcr.ID(*cmd.pcrIndex), tpmeventlog.TPMAlgorithm(*cmd.hashAlgo), nil)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unable to replay the PCR%d value: %v", *cmd.pcrIndex, err)
+			return
+		}
+		fmt.Printf("Calc\t%2d\t%10s\t%3d\t%X\t\n", *cmd.pcrIndex, "", *cmd.hashAlgo, calculatedValue)
+	}
+}
```
ChriMarMe, line 92: What is the reason behind not returning the error and printing it on os.Stderr instead?

cmd/pcr0tool/commands/displayfwinfo/command.go
```diff
@@ -0,0 +1,75 @@
...
+		return
+	}
+
+	b, err := json.Marshal(dmiTable.BIOSInfo())
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to serialize BIOSInfo: %v", err)
+		return
+	}
+	fmt.Printf("%s\n", b)
+}
```
ChriMarMe, line 75: Same question here: 
What is the reason behind not returning the error and printing it on os.Stderr instead?

cmd/pcr0tool/commands/dumpregisters/helpers/get_registers.go
```diff
@@ -0,0 +1,98 @@
...
+			return cfg, fmt.Errorf("unable to apply option %T:%v due to error: %w", opt, opt, err)
+		}
+	}
+	return cfg, nil
+}
+
+// GetRegisters extracts status registers.
+func GetRegisters(
+	options ...GetRegistersOption,
+) (registers.Registers, error) {
```
ChriMarMe, line 57: Can this be in one line? I dont see a mass of arugments passed that could lead to confusion, but maybe you have your reasons and want to share? :)

pkg/bruteforcer/indexes.go
```diff
@@ -0,0 +1,333 @@
...
+		l := len(s)
+		for i := idx; i < l; i, newValue = i+1, newValue+1 {
+			s[i] = newValue
+		}
+		if s[l-1] > maxValue {
+			panic(fmt.Sprintf("internal error: should never happen: value %d is greater than maxValue %d", s[l-1], maxValue))
+		}
+	}
+
+	// TODO: check maxValue and combinationID sanity.
```
ChriMarMe, line 121: Is this still relevant?

pkg/pcrbruteforcer/reproduce_expected_pcr0.go
```diff
@@ -0,0 +1,705 @@
...
+
+func (j *reproduceExpectedPCR0Job) tryDisabledMeasurementsCombination(
+	ctx xcontext.Context,
+	disabledMeasurementsCombination bruteforcer.UniqueUnorderedCombination,
+	hashFuncFactory func() hash.Hash,
+) (bool, *registers.ACMPolicyStatus, error) {
+	var enabledMeasurements []*pcr.CachedMeasurement
+	for idx, m := range j.measurements {
+		isDisabled := false
+		// TODO: refactor this O(N^2); generate the picked measurements from the get go
```
ChriMarMe, line 331: Is this ToDo still relevant?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func getFirmware() *uefi.UEFI {
+	// It appears Y3DLD11 has wrong order of elements inside BPM:
+	// PFRE goes PCDE, thus disabling the host strict order check.
+	//
+	// TODO: make offsets stable to elements reordering, currently
+	// if position of PMSE will be wrong it will show a wrong area
+	// in the report.
```
ChriMarMe, line 27: Is this ToDo still relevant?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
```diff
@@ -0,0 +1,181 @@
...
+	manifest.StrictOrderCheck = false
+
+	firmwareImage, err := decompressImage("firmwares", "F0E_3A10.bin.xz")
+	if err != nil {
+		panic(err)
+	}
```
ChriMarMe, line 33: Is this worth a panic?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	firmware, err := uefi.ParseUEFIFirmwareBytes(firmwareImage)
+	if err != nil {
+		panic(err)
+	}
```
ChriMarMe, line 38: Is this worth a panic as well?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
```diff
@@ -0,0 +1,181 @@
...
+func BenchmarkReproduceExpectedPCR0(b *testing.B) {
+	firmware := getFirmware()
+
+	const correctACMRegValue = 0x0000000200108681
+	ctx := xcontext.Background()
+
+	pcr0Correct, err := hex.DecodeString("15648D5FFD864AFC9FFB66D7FCE133F089F77A45")
+	if err != nil {
+		panic(err)
+	}
```
ChriMarMe, line 106: Why all the panic?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	pcr0Incomplete, err := hex.DecodeString("E9C988C47137770B0F26C8900B1D6761B95199F7")
+	if err != nil {
+		panic(err)
+	}
+
+	pcr0Invalid, err := hex.DecodeString("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
+	if err != nil {
+		panic(err)
+	}
```
ChriMarMe, line 116: What is this panic all about?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
```diff
@@ -0,0 +1,181 @@
...
+				pcr.SetFlow(pcr.FlowIntelCBnT0T),
+				pcr.SetIBBHashDigest(tpm2.AlgSHA1),
+				pcr.SetRegisters(registers.Registers{
+					registers.ParseACMPolicyStatusRegister(correctACMRegValue + acmCorruption),
+				}),
+			}
+			measurements, _, _, err := pcr.GetMeasurements(firmware, 0, measureOptions...)
+			if err != nil {
+				panic(err)
+			}
```
ChriMarMe, line 136: Pan!c at the disco.

ChriMarMe (commented):
Some nits, but nothing serious. :) 

Good work. :+1:

cmd/pcr0tool/commands/displayeventlog/command.go
xaionaro: Not really. Just a legacy. I'm just continuing the same patterns used in other parts of the pcr0tool code :(

Let's agree this is a tech debt for another PR (to rework the whole code)?

pkg/pcrbruteforcer/reproduce_expected_pcr0.go
xaionaro: Yeah, relevant, but not the priorirty

pkg/bruteforcer/indexes.go
xaionaro: Relevant, but not the priority.

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
xaionaro: I would'be pass `*testing.T`, but this also used for benchmarks, so what way do we have to terminate an error?

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
xaionaro: Fixed.

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
xaionaro: Fixed.

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
xaionaro: Fixed.

cmd/pcr0tool/commands/displayfwinfo/command.go
xaionaro: Answered in: https://github.com/9elements/converged-security-suite/pull/325#discussion_r854133968

cmd/pcr0tool/commands/dumpregisters/helpers/get_registers.go
xaionaro: Answered in: https://github.com/9elements/converged-security-suite/pull/325#discussion_r854133968

pkg/pcrbruteforcer/reproduce_expected_pcr0_test.go
xaionaro: Fixed.

cmd/pcr0tool/commands/displayeventlog/command.go
ChriMarMe: Fine with me :)

cmd/pcr0tool/commands/displayfwinfo/command.go
ChriMarMe: jip

ChriMarMe (approved):

pkg/registers/registers.go
```diff
@@ -57,42 +70,138 @@ func (regs Registers) Find(id RegisterID) Register {
...
+func (regs *Registers) UnmarshalYAML(value *yaml.Node) error {
+	var regsYaml registersForSerialization
+	if err := value.Decode(&regsYaml); err != nil {
+		return fmt.Errorf("unable to un-YAML-ize registers: %w", err)
+	}
+	result, err := regsYaml.Registers()
+	if err != nil {
+		return fmt.Errorf("unable to parse registers data: %w", err)
+	}
+	*regs = result
```
xaionaro, line 205: Now the file with registers looks this way:
```
ACM_POLICY_STATUS: 0x12000d8e1
ACM_STATUS: 0xc0008000
TXT.DIDVID: 0x1b00c8086
TXT.DPR: 0x0
TXT.ERRORCODE: 0x0
TXT.ESTS: 0x0
TXT.HEAP.BASE: 0x0
TXT.HEAP.SIZE: 0x0
TXT.MLE.JOIN: 0x0
TXT.PUBLIC.KEY: 0x879a8f9cbf9e3d1d12dc9ad76dde34e6aa403664c739db347b858f0be033ae3a
TXT.SINIT.BASE: 0x0
TXT.SINIT.SIZE: 0x0
TXT.SPAD: 0x9800000080000000
TXT.STS: 0x12
TXT.VER.EMIF: 0x9d003000
TXT.VER.FSBIF: 0xffffffff
```
instead of old format:
```
[{"id":"ACM_POLICY_STATUS","value":"5dgAIAEAAAA="},{"id":"ACM_STATUS","value":"AIAAwAAAAAA="},{"id":"TXT.DPR","value":"AAAAAA=="},{"id":"TXT.ERRORCODE","value":"AAAAAA=="},{"id":"TXT.PUBLIC.KEY","value":"h5qPnL+ePR0S3JrXbd405qpANmTHOds0e4WPC+Azrjo="},{"id":"TXT.STS","value":"EgAAAAAAAAA="},{"id":"TXT.ESTS","value":"AA=="},{"id":"TXT.SPAD","value":"AAAAgAAAAJg="},{"id":"TXT.VER.FSBIF","value":"/////w=="},{"id":"TXT.VER.EMIF","value":"ADAAnQ=="},{"id":"TXT.DIDVID","value":"hoAMsAEAAAA="},{"id":"TXT.SINIT.BASE","value":"AAAAAA=="},{"id":"TXT.SINIT.SIZE","value":"AAAAAA=="},{"id":"TXT.MLE.JOIN","value":"AAAAAA=="},{"id":"TXT.HEAP.BASE","value":"AAAAAA=="},{"id":"TXT.HEAP.SIZE","value":"AAAAAA=="}]
```

rihter007: Marhsalling/unmarshalling registers from specific format should not be a method of Registers. Please make it as a funciton.

pkg/registers/registry.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// AddRegister remembers the register type and ID.
+func (registry registryT) AddRegister(reg Register) {
+	registry.idToType[reg.ID()] = reflect.TypeOf(reg)
+}
+
+// New constructs a register, given its ID and raw value. It
+// will return an error if the value is of wrong underlying type.
+func (registry registryT) New(regID RegisterID, value interface{}) (Register, error) {
```
rihter007, line 23: Why do we need it?
If one wants a register X, it can can call NewRegisterX and pass specific arguments instead of interface{}

pkg/registers/registers.go
rihter007: Previous MarshalJSON should also be implemented as a function :)

pkg/registers/registry.go
xaionaro: This is used in `UnmarshalYAML`, which has a dynamic nature.

pkg/registers/registers.go
xaionaro: My logic is:
1. This is an intrinsic logic of `Registers`, why it should be defined outside of type `Registers`?
2. It is an usual practice in Go to define a method in such cases. I guess in all project I've seen it was defined as methods, not as functions.
3. When you define this as the method, you satisfy the [`Marshaler`](https://pkg.go.dev/encoding/json#Marshaler) and [`Unmarshaler`](https://pkg.go.dev/encoding/json#Unmarshaler) interfaces, which enables you to just use for example `json.Marshal` either directly on `Registers` or even on structure which includes `Registers`.

Could please you elaborate why it shouldn't be a method?

pkg/registers/marshalling.go
```diff
@@ -150,13 +154,14 @@ func getRegister8Parser(regID RegisterID) func(ui8 uint8) Register {
 	return nil
 }
 
-// Unmarshal constructs register from it's id and marshalled value
-func Unmarshal(id RegisterID, b []byte) (Register, error) {
+// ValueFromBytes constructs register from it's id and marshalled value
+// TODO: use dynamic conversion instead of hard-code
+func ValueFromBytes(id RegisterID, b []byte) (Register, error) {
```
rihter007, line 159: nit RegisterFromBytes.

The counter call was named Marshal"Value" as it marshalled only the value of the register

pkg/registers/registers.go
rihter007: >> Could please you elaborate why it shouldn't be a method?

Because without a method it could be even implemented in a different package.
A method introduces extra coupling that should be avoided if possible.

I forgot that golang requires types to implement interfaces, so it is ok to keep it as a method in this case

pkg/registers/registry.go
rihter007: UnmarshalJSON didn't require to have a registry for that purpose.

I saw your comments about being able to "dynamically" parse registers.
IMHO it looks like hardcoding of adding a bit of lines to "dynamically" add register to registry vs hardcoded in switch case by id.

I'm ok to go with registry, but would prefer a bit different strategy

pkg/registers/registers.go
```diff
@@ -57,42 +70,138 @@ func (regs Registers) Find(id RegisterID) Register {
...
 		})
 	}
-	return json.Marshal(resultRegisters)
+	return resultRegisters, nil
+}
+
+// MarshalJSON converts a collection of registers into a JSON
+// TODO: remove this, obsolete format
+// OBSOLETE
+func (regs Registers) MarshalJSON() ([]byte, error) {
```
rihter007, line 114: JSON itself is not an obsolete format. I am ok to add new.

What if someone has a structure it wants to serialise in json and Registers is a field in that structure?

pkg/registers/registers.go
```diff
@@ -11,6 +17,7 @@ type Register interface {
 	BitSize() uint8
 	Fields() []Field
 	Address() uint64
+	Value() interface{}
```
rihter007, line 20: why?

If someone wants a specific value of specific register, it can call

reg = FindRegisterX(registers)
reg.Raw() - which has specific type

For generic purpose there are Marshal/Unmarshal methods.

I'm ok to implement something better than what we have right now in marshal/unmarshal methods.

Having a switch of RawRegister in Marshal is not hardcoding in my view as it is done once.

Having an extra public method for that single purpose makes little sense to me. As every customer will have to guess what is that interface{} (same as with RawRegister)

pkg/registers/registers.go
xaionaro: I actually believe we should use a map instead of an array in JSON as well. I kept this `MarshalJSON` only for backward compatibility (to be able to read `registers.json` dumped earlier). Would it be fine if I will do next:
* Will reimplement `MarshalJSON` to save it as map.
* Will implement a backward-compatible parser as a separate helper.

rihter007: Yes, this sounds reasonable! Thanks :)

cmd/cbnt-prov/cmd.go
```diff
@@ -499,7 +499,7 @@ func (g *generateBPMCmd) Run(ctx *context) error {
 
 	bpm, err := cbnt.GenerateBPM(options, g.BIOS)
 	if err != nil {
-		return err
+		return fmt.Errorf("GenerateBPM: %w", err)
```
rihter007, line 502: nit
AFAIK error messages should start with lower case

cmd/pcr0tool/commands/dumpregisters/helpers/get_registers.go
```diff
@@ -0,0 +1,96 @@
...
+func GetRegisters(options ...GetRegistersOption) (registers.Registers, error) {
+	cfg, err := getRegistersOptions(options).Config()
+	if err != nil {
+		return nil, fmt.Errorf("invalid arguments: %w", err)
+	}
+	if runtime.GOOS != "linux" {
+		return nil, fmt.Errorf("command is supported only on Linux platform")
+	}
+
+	// TODO: add support for non-Intel systems
```
rihter007, line 64: nit

It would be nice to have a graceful check for non-intel

cmd/pcr0tool/commands/dumpregisters/helpers/get_registers.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("TXT public space is not provided: %w", err)
+	}
+
+	msrReader := &registers.DefaultMSRReader{}
+	txtRegisters, txtErr := registers.ReadTXTRegisters(txtConfig)
+	msrRegisters, msrErr := registers.ReadMSRRegisters(msrReader)
+	allRegisters := append(txtRegisters, msrRegisters...)
+
+	return allRegisters, (&errors.MultiError{}).Add(txtErr, msrErr).ReturnValue()
```
rihter007, line 75: nit

IMHO situation when both registers and an error exist is confusing... (I personally made a bug because of that). Unfortunately I do not know what is the best way.

cmd/pcr0tool/commands/sum/command.go
```diff
@@ -83,6 +92,21 @@ func (cmd Command) Execute(args []string) {
 		usageAndExit()
 	}
 
+	if *cmd.decrementACMPolicyStatus != 0 {
+		found := false
```
rihter007, line 96: nit

var found bool

pkg/bruteforcer/brute_forcer.go
```diff
@@ -0,0 +1,187 @@
...
+				for i := uint64(0); ; {
+					if resultData != nil {
+						// we do not need atomic in the condition above, because
+						// few false negatives will not affect anything much.
+						return
+					}
+
+					if b.try(bfctx, dataCopy, iterator) {
+						locker.Lock()
+						resultData = iterator.GetCombination()
```
rihter007, line 141: nit

it would be nice to notify other goroutines to stop

pkg/pcrbruteforcer/reproduce_event_log.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	pcr0Data := m.CompileMeasurableData(imageBytes)
+
+	// maxDistance is the maximal hamming distance to brute force to, it is
+	// picked to 6, because it is maximal value for a reasonable brute-force
+	// time (about 1 second).
+	//
+	// For benchmark details see also README.md of package `bruteforcer`.
+	maxDistance := 6
```
rihter007, line 164: nit

it looks like it should be configurable (for tools).
+ a good default value should be put into constants

pkg/pcrbruteforcer/reproduce_event_log.go
```diff
@@ -0,0 +1,195 @@
...
+
+	newContextFunc := func() (interface{}, error) {
+		buf := make([]byte, len(pcr0Data))
+		copy(buf, pcr0Data)
+		return &bruteForceACMPolicyStatusEventContext{
+			Hasher: sha1.New(),
```
rihter007, line 170: nit

Why only sha1?

rihter007 (approved):
All in all looks good!
Left some code-style nits

pkg/pcrbruteforcer/reproduce_event_log.go
xaionaro: `SHA1` is faster to brute-force, but yeah, this is a lack of functionality. Basically here I just copied the code from the internal tooling without any change.

pkg/bruteforcer/brute_forcer.go
xaionaro: We do it above in:
```
					if resultData != nil {
						// we do not need atomic in the condition above, because
						// few false negatives will not affect anything much.
						return
					}
```
In the worst case scenario other routines would do like 2 extra iterations.

cmd/pcr0tool/commands/dumpregisters/helpers/get_registers.go
xaionaro: Good point.

I guess, we need to separate warnings from errors in the return. Let me add a TODO about that.

pkg/registers/registers.go
xaionaro: I kinda just push this into debt to be solved when we will start using generics. We'll be able to define a generic set of methods for a generic register, and it would avoid hardcoding these for each register.

I agree, it is arguable, but what do you think?

xaionaro:
> All in all looks good! Left some code-style nits

Thank you for the input. I'm a bit afraid to reset the approval by fixes, so I'll make a dedicated PR for the most of the suggestions, since the suggestions are non-blocking. Hoping it's fine.

zaolin (approved):

xaionaro:
@zaolin : For some reason it still does not allow to merge :(

walterchris (approved):

PR #330, opened 2022-05-26
Add a skeleton of redesigned "pcr" package

xaionaro:
Adding the main entities of a new design of PCR values calculations process. New package is called "bootflow" (instead of "pcr"), since it is not longer focused on PCR values only. It also supports other roots of trust for storage and measurements.

New design (in contrast to the old one) is:
* Extensible from external packages.
* TPM terms "Event", "Extend" and "Measurement" are clarified.
* Flow now consists of steps. Each step may contain multiple measurements. It solves the problem with dynamic amount of measurements on some platforms.
* New design also aware of DICE and other chains of trust.

uml

UML is generated by command:
```
goplantuml -recursive pkg/bootflow/ | sed -e 's/\#\.\./\*--/g' | sed -re 's/\[\]([^b])/\1/g'
```

An example of a flow description:
https://github.com/9elements/converged-security-suite/pull/330/files#diff-27af1a8ff010eb439ee11da802024b19fbf9bf11ea6580efcbfed47f54135c89R9-R15

### DIFF
--- a/pkg/bootflow/actions/commonactions/set_flow.go
+++ b/pkg/bootflow/actions/commonactions/set_flow.go
@@ -1,0 +1,23 @@
+package commonactions
+
+import (
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/types"
+)
+
+type setFlow struct {
+	nextFlowFunc   func() types.Flow
+	startStepIndex uint
+}
+
+func SetFlow(flowFunc func() types.Flow, stepIndex uint) types.Action {
+	return &setFlow{
+		nextFlowFunc:   flowFunc,
+		startStepIndex: stepIndex,
+	}
+}
+
+func (step *setFlow) Apply(state *types.State) error {
+	state.CurrentFlow = step.nextFlowFunc()
+	state.CurrentStepIdx = step.startStepIndex
+	return nil
+}
--- a/pkg/bootflow/bootengine/boot_process.go
+++ b/pkg/bootflow/bootengine/boot_process.go
@@ -1,0 +1,78 @@
+package bootengine
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/types"
+)
+
+type BootProcess struct {
+	// Do not mutate these values from the outside, the mutability
+	// is owned by BootProcess itself.
+	CurrentState *types.State
+	Log          Log
+}
+
+func NewBootProcess(state *types.State) *BootProcess {
+	return &BootProcess{
+		CurrentState: state,
+	}
+}
+
+func stateNextStep(state *types.State) (types.Step, StepIssues, bool) {
+	if state.CurrentFlow == nil {
+		return nil, nil, false
+	}
+
+	if state.CurrentStepIdx >= uint(len(state.CurrentFlow)) {
+		return nil, nil, false
+	}
+
+	step := state.CurrentFlow[state.CurrentStepIdx]
+	actions := step.Actions(state)
+	var stepIssues StepIssues
+	for idx, action := range actions {
+		issue := action.Apply(state)
+		if issue != nil {
+			stepIssues = append(stepIssues, StepIssue{ActionIndex: uint(idx), Issue: issue})
+		}
+	}
+
+	state.CurrentStepIdx++
+	return step, stepIssues, true
+}
+
+func (process *BootProcess) NextStep() bool {
+	oldVerifiedData := process.CurrentState.VerifiedData
+	stepBackend, stepIssues, ok := stateNextStep(process.CurrentState)
+	if !ok {
+		return false
+	}
+	step := Step{Step: stepBackend}
+	step.Issues = stepIssues
+
+	if len(process.CurrentState.VerifiedData) > len(oldVerifiedData) {
+		step.VerifiedData = process.CurrentState.VerifiedData[len(oldVerifiedData):]
+	}
+
+	process.Log = append(process.Log, step)
+	return true
+}
+
+func (process *BootProcess) Finish() {
+	for process.NextStep() {
+	}
+}
+
+func (process *BootProcess) GoString() string {
+	var result strings.Builder
+	fmt.Fprintf(&result, "Current state:\n\t%s\n", nestedGoStringOf(process.CurrentState))
+	fmt.Fprintf(&result, "Resulting steps:\n\t%s\n", nestedGoStringOf(process.Log))
+	return result.String()
+}
+
+func nestedGoStringOf(i interface{}) string {
+	v := fmt.Sprintf("%#v", i)
+	return strings.ReplaceAll(strings.Trim(v, "\n"), "\n", "\n\t")
+}
--- a/pkg/bootflow/bootengine/boot_process.go
+++ b/pkg/bootflow/bootengine/boot_process.go
@@ -20,13 +20,13 @@ func NewBootProcess(state *types.State) *BootProcess {
 	}
 }
 
-func stateNextStep(state *types.State) (types.Step, StepIssues, bool) {
+func stateNextStep(state *types.State) (types.Step, types.Actions, StepIssues, bool) {
 	if state.CurrentFlow == nil {
-		return nil, nil, false
+		return nil, nil, nil, false
 	}
 
 	if state.CurrentStepIdx >= uint(len(state.CurrentFlow)) {
-		return nil, nil, false
+		return nil, nil, nil, false
 	}
 
 	step := state.CurrentFlow[state.CurrentStepIdx]
@@ -40,16 +40,16 @@ func stateNextStep(state *types.State) (types.Step, StepIssues, bool) {
 	}
 
 	state.CurrentStepIdx++
-	return step, stepIssues, true
+	return step, actions, stepIssues, true
 }
 
 func (process *BootProcess) NextStep() bool {
 	oldVerifiedData := process.CurrentState.VerifiedData
-	stepBackend, stepIssues, ok := stateNextStep(process.CurrentState)
+	stepBackend, actions, stepIssues, ok := stateNextStep(process.CurrentState)
 	if !ok {
 		return false
 	}
-	step := Step{Step: stepBackend}
+	step := Step{Step: stepBackend, Actions: actions}
 	step.Issues = stepIssues
 
 	if len(process.CurrentState.VerifiedData) > len(oldVerifiedData) {
--- a/pkg/bootflow/examples/bruteforce/main.go
+++ b/pkg/bootflow/examples/bruteforce/main.go
@@ -1,0 +1,131 @@
+package main
+
+import (
+	"bytes"
+	"crypto/sha1"
+	"fmt"
+	"hash"
+
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/bootengine"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/steps/tpmsteps"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/trustchains/tpm"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/types"
+	"github.com/9elements/converged-security-suite/v2/pkg/bruteforcer"
+	"github.com/google/go-tpm/tpm2"
+)
+
+func main() {
+	// Here we brute-force a TPM measurement against a final PCR0 value.
+	//
+	// We initially have measurements of two separators, but we need to
+	// bruteforce the first measurement to 0x01000001
+
+	// an artificial flow for two separators
+	myFlow := types.Flow{
+		tpmsteps.InitTPMLazy(0),
+		tpmsteps.MeasureSeparator(0),
+		tpmsteps.MeasureSeparator(0),
+	}
+
+	// the expected final hash (after bruteforcing)
+	expectedHash := make([]byte, sha1.Size)
+	extend := func(h hash.Hash, a, b []byte) []byte {
+		defer h.Reset()
+		h.Write(a)
+		h.Write(b)
+		return h.Sum(nil)
+	}
+	sep0, sep1 := sha1.Sum([]byte("\x01\x00\x00\x01")), sha1.Sum([]byte("\x00\x00\x00\x00"))
+	expectedHash = extend(sha1.New(), expectedHash, sep0[:])
+	expectedHash = extend(sha1.New(), expectedHash, sep1[:])
+
+	// executing the flow (with two simple "\x00\x00\x00\x00" measurements)
+	state := types.NewState()
+	state.SetFlow(myFlow, 0)
+	state.EnableTrustChain(tpm.NewTPM())
+	process := bootengine.NewBootProcess(state)
+	process.Finish()
+
+	// just some debugging
+	fmt.Printf("Log:\n%#v\n", process.Log)
+
+	// == now let's bruteforce the first measurement ==
+
+	// first find the original (UNHASHED) data of the first measurement
+	tpmMeasurements := process.Log.GetVerifiedDataByTrustChainType((*tpm.TPM)(nil))
+
+	// now let's gather the rest chain (TPM init and the rest TPM Extend-s)
+	var (
+		tpmLocality *uint8
+		tpmExtends  []tpm.CommandLogEntryExtend
+	)
+	tpm.StateExec(state, func(t *tpm.TPM) error {
+		for _, entry := range t.CommandLog {
+			switch entry := entry.(type) {
+			case tpm.CommandLogEntryInit:
+				tpmLocality = &[]uint8{entry.Locality}[0]
+			case tpm.CommandLogEntryExtend:
+				if entry.HashAlgo != tpm2.AlgSHA1 {
+					continue
+				}
+				tpmExtends = append(tpmExtends, entry)
+			}
+		}
+		return nil
+	})
+	if tpmLocality == nil {
+		panic("the TPMInit command wasn't found")
+	}
+
+	// let's construct the final flow:
+
+	type context struct {
+		sha1Hasher hash.Hash
+	}
+	combination, err := bruteforcer.BruteForceBytes(
+		tpmMeasurements[0].Data.ForceBytes,
+		2,
+		func() (interface{}, error) {
+			return &context{
+				sha1Hasher: sha1.New(),
+			}, nil
+		},
+		func(_ctx interface{}, data []byte) bool {
+			ctx := _ctx.(*context)
+			h := ctx.sha1Hasher
+
+			// starting a PCR value from scratch:
+			pcrValue := make([]byte, sha1.Size)
+			pcrValue[len(pcrValue)-1] = *tpmLocality
+
+			// hashing the bruteforced value before extending
+			h.Write(data)
+			dataHashed := h.Sum(nil)
+			h.Reset()
+
+			// extending it
+			pcrValue = extend(h, pcrValue, dataHashed)
+
+			// extending the rest of the measurements (they are already pre-hashed)
+			for _, tpmExtend := range tpmExtends[1:] {
+				pcrValue = extend(h, pcrValue, tpmExtend.Digest)
+			}
+
+			// is it OK?
+			return bytes.Equal(pcrValue, expectedHash)
+		},
+		0,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	// printing the result
+	result := []byte("\x00\x00\x00\x00")
+	combination.ApplyBitFlips(result)
+	fmt.Printf("COMBINATION: %#v\n", combination)
+	fmt.Printf("RESULT: 0x%X\n", result)
+
+	tpmMeasurements[0].Data.ForceBytes = result
+	fmt.Printf("resulting measurements: %#v\n", tpmMeasurements)
+}
--- a/pkg/bootflow/examples/bruteforce/main.go
+++ b/pkg/bootflow/examples/bruteforce/main.go
@@ -77,7 +77,7 @@ func main() {
 		panic("the TPMInit command wasn't found")
 	}
 
-	// let's construct the final flow:
+	// brute force:
 
 	type context struct {
 		sha1Hasher hash.Hash
--- a/pkg/bootflow/examples/ocpdxepcr0/main.go
+++ b/pkg/bootflow/examples/ocpdxepcr0/main.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/bootengine"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/flows"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/systemartifacts/biosfirmware"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/trustchains/tpm"
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/types"
+	"github.com/google/go-tpm/tpm2"
+)
+
+func main() {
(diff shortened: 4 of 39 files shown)


pkg/bootflow/trustchains/trust_chain_tpm.go
```diff
@@ -0,0 +1,108 @@
...
+var _ types.TrustChain = (*TPM)(nil)
+
+type TPM struct {
+	PCRValues PCRValues
+	EventLog  TPMEventLog
+}
+
+func TPMFunc(state *types.State, fn func(tpm *TPM) error) error {
+	for _, trustChain := range state.TrustChains {
+		if tpm, ok := trustChain.(*TPM); ok {
```
mimir-d, line 20: the fact that you need to check rt type info here seems like a bit of code smell to me. you could probably keep a map of known chains and avoid this func completely; it wouldnt avoid extensibility, although realistically speaking, it's not every day that a new trust mechanism is added

mimir-d: i see the same pattern happens with state.SystemArtifacts, and you need to do a linear search every time there as well

pkg/bootflow/bootengine/boot_process.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/9elements/converged-security-suite/v2/pkg/bootflow/types"
+)
+
+type BootProcess struct {
+	// Do not mutate these values from the outside, the mutability
```
mimir-d, line 11: you can use an accessor func here for guaranteed immutability

pkg/bootflow/examples/ocpdxepcr0/main.go
```diff
@@ -0,0 +1,57 @@
...
+	// the main part
+	state := types.NewState()
+	state.SetFlow(flows.OCPDXE(), 0)
+	state.EnableTrustChain(tpm.NewTPM())
+	state.EnableSystemArtifact(biosfirmware.NewBIOSFirmware(biosFirmware))
+	process := bootengine.NewBootProcess(state)
+	process.Finish()
+
+	// printing results
+	tpm.StateExec(state, func(t *tpm.TPM) error {
```
mimir-d, line 34: design nit; are there any alternatives to using this state object as an opaque pointer? this reads a bit weird to me, although it does address extensibility (which i assume was the main reason for this form in the first place)

pkg/bootflow/examples/bruteforce/main.go
```diff
@@ -0,0 +1,131 @@
...
+	process := bootengine.NewBootProcess(state)
+	process.Finish()
+
+	// just some debugging
+	fmt.Printf("Log:\n%#v\n", process.Log)
+
+	// == now let's bruteforce the first measurement ==
+
+	// first find the original (UNHASHED) data of the first measurement
+	tpmMeasurements := process.Log.GetVerifiedDataByTrustChainType((*tpm.TPM)(nil))
```
mimir-d, line 55: design nit; this typed nil (although valid syntax) may look weird for less experienced. is there a way to express extensibility (external i suppose in this case) without using the type system like this? id also be interested in understanding more regarding the extensibility aspect (as i see it used a lot in this whole design) and why you'd want to encode it like this (rather than, for example, leave consumers to derive their own code or contrib to this codebase directly)

pkg/bootflow/examples/bruteforce/main.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// now let's gather the rest chain (TPM init and the rest TPM Extend-s)
+	var (
+		tpmLocality *uint8
+		tpmExtends  []tpm.CommandLogEntryExtend
+	)
+	tpm.StateExec(state, func(t *tpm.TPM) error {
```
mimir-d, line 62: this linear search here to extract individual parts from the measure chain seems brittle to me and is different from the design where we had a list of tpm measurements in the sense that, in that case, we could just use the list directly without having knowledge of the flow and alter one item from it in the bruteforcer. In this case, it's basically required to recreate the flow in code (lower down in the bruterforce func) to do the same thing

pkg/bootflow/types/state.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func NewState() *State {
+	return &State{
+		SystemArtifacts: map[reflect.Type]SystemArtifact{},
+		TrustChains:     map[reflect.Type]TrustChain{},
+	}
+}
+
+func (state *State) SetFlow(flow Flow, stepIdx uint) {
```
marcoguerri, line 36: Does the state need to be rolled back when transitioning to new flow? I think we agree we it's hard to guarantee it's done correctly. Should the boot process restart from a clean state (while keeping track of the encountered issues), when a `Flow` fails and we need to switch? Heavy operations could be cached.

pkg/bootflow/bootengine/boot_process.go
xaionaro: Yeah, but didn't want to decide how exactly I will handle this, yet.

pkg/bootflow/examples/bruteforce/main.go
xaionaro: > design nit; this typed nil (although valid syntax) may look weird for less experienced.

It looks weird even for myself :D

> is there a way to express extensibility (external i suppose in this case) without using the type system like this? 

Well. It does not require a `nil`. One may call for example:
```
tpmMeasurements := process.Log.GetVerifiedDataByTrustChainType(tpm.TPM{})
```
instead, and it still will work.

Or we can select not by type, but by a specific instance:
```
tpmInstance := tpm.NewTPM()
state.EnableTrustChain(tpmInstance)
[...]
tpmMeasurements := process.Log.GetVerifiedDataByTrustChain(tpmInstance)
```
would that be better?

pkg/bootflow/examples/bruteforce/main.go
xaionaro: > in that case, we could just use the list directly without having knowledge of the flow and alter one item from it in the bruteforcer.

Here is the same. We do not know the flow, we cherry-pick the first measurement (the same as earlier) and brute force it. The differences are:
1. We manually set TPM locality, but again we extract it from the flow (and we do not need to know the flow to do that).
2. We manually do the "Extend" chain instead of calling `Calculate`. In other words, the internals of function `Calculate` are partly reimplemented in the brute-forcer.

But the brute-forcer is agnostic to the specific flow: the only requirement is to have at least the one measurement, because we brute-force it -- but it was the case in the old design as well.

pkg/bootflow/examples/ocpdxepcr0/main.go
xaionaro: Could you rephrase, please?

Do I understand correctly the question is: can we avoid this `StateExec` function? If so, the answer is yes:
If above we would do:
```
tpmInstance := tpm.NewTPM()
state.EnableTrustChain(tpmInstance)
```
instead of:
```
state.EnableTrustChain(tpm.NewTPM())
```
then we could've just access `tpmInstance` directly without calling `StateExec`.

pkg/bootflow/actions/commonactions/set_flow.go
```diff
@@ -0,0 +1,23 @@
...
+
+type setFlow struct {
+	nextFlowFunc   func() types.Flow
+	startStepIndex uint
+}
+
+func SetFlow(flowFunc func() types.Flow, stepIndex uint) types.Action {
+	return &setFlow{
+		nextFlowFunc:   flowFunc,
+		startStepIndex: stepIndex,
```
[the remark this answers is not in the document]
xaionaro: OK, good point. I added this only for extra flexibility. Just haven't thought of this.

Yeah, to remove `stepIndex` is easy (actually easier than preserve it).

pkg/bootflow/bootengine/boot_process.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, nil, nil, false
+	}
+
+	step := state.CurrentFlow[state.CurrentStepIdx]
+	actions := step.Actions(state)
+	var stepIssues StepIssues
+	for idx, action := range actions {
+		issue := action.Apply(state)
+		if issue != nil {
+			stepIssues = append(stepIssues, StepIssue{ActionIndex: uint(idx), Issue: issue})
```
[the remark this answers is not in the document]
xaionaro: > Also `ActionIndex` tracking for issues seems to be very fragile.

`ActionIndex` is supposed to be used only for debugging. The specific Issue should be defined by error type returned by `action.Apply`.

pkg/bootflow/types/state.go
xaionaro: > Does the state need to be rolled back when transitioning to new flow?

In the current proposal we do not roll back.

> Should the boot process restart from a clean state (while keeping track of the encountered issues), when a `Flow` fails and we need to switch?

In the current proposal we are not restarting the boot process.

---

The `SetFlow` is like `JMP`, not like `CALL`. There is no returning to the original `Flow`, no saving state. Just sequentially execute steps until there will be no more steps to execute.

Basically a `Flow` might be of the of, for example (in pseudo-code):
```
Root {
    IfFITPresent{SetFlow{Intel}}
    IfAMDManifestPresent{SetFlow{AMD}}
}

Intel {
    IfBPMPresent{SetFlow{CBnT}}
    SetFlow{LegacyTXT}
}

CBnT {
    ValidateACM{Failure: SetFlow{CBnTFailure}}
    ValidateKM{Failure: SetFlow{CBnTFailure}}
    ValidateBPM{Failure: SetFlow{CBnTFailure}}
    ValidateIBB{Failure: SetFlow{CBnTFailure}}
    TPMInit(3)
    MeasurePCR0DATA
    SetFlow{IntelResetVector}
}

CBnTFailure {
    IfFITAddrIs{specialFitType, specialAddrValue, CBnT0TFailureMeta} // check if a special FIT entry has a special value
    SetFlow{IntelResetVector}
}

CBnT0TFailureMeta {
    MeasurePCR0DATAFailure
    SetFlow{IntelResetVector}
}

IntelResetVector {
    IfOCP{SetFlow{OCPPEI}}
    IfHPE{SetFlow{HPEPEI}}
    IfLenovo{SetFlow{LenovoPEI}}
}

OCPPEI {
    TPMInitLazy(3)
    MeasureUEFIGUID(0, OCP_PEI_GUID)
    MeasureSeparator(0)
    SetFlow{OCPDXE}
}

OCPDXE {
    MeasurePCDVariable(0, "FirmwareVendorVersion")
    MeasureUEFIGUID(0, OCP_DXE_GUID)
    MeasureSeparator
}
```

marcoguerri: Ok, as long as the contract is clear that a flow should be built to leave consistent state behind. Hard to enforce, but this is what should happen in real world. It's not always the case, but this doesn't mean we should do it right here :) 

As an example, Intel ACM I think behaves slightly differently (e.g. requirement to cap PCR registers after initializing TPM if NVRAM indexes are provisioned in wrong way). Intel ACM would actually have:

```
    TPMInit(3)
    ValidateACM{Failure: SetFlow{CBnTFailure}}
    ValidateKM{Failure: SetFlow{CBnTFailure}}
    ValidateBPM{Failure: SetFlow{CBnTFailure}}
    ValidateIBB{Failure: SetFlow{CBnTFailure}}
    MeasurePCR0DATA
    SetFlow{IntelResetVector}
```

But ok, Intel ACM is the broken one, no need to replicate broken behavior :)

pkg/bootflow/trustchains/trust_chain_tpm.go
xaionaro: Fixed. Now we use a map instead

pkg/bootflow/bootengine/boot_process.go
mimir-d: ok, just making sure it's not left as a "temporary todo" :D

pkg/bootflow/examples/ocpdxepcr0/main.go
mimir-d: oh... so that's the semantic. Wasnt clear to me from the naming "StateExec". Well, why not have a "GetChain" instead of the lambda call? What's the benefit here? (i think you have a similar pattern for Log)

pkg/bootflow/examples/bruteforce/main.go
mimir-d: I think this is an artifact of the design. You're essentially using the type system as an enum value here (and discarding the actual instance value, which here is nil, or even worse tpm.TPM{}). It's usually more clear to be direct about it (as in, use an actual enum), but then again, I don't have a simple localized proposal for this right now considering all the other code

pkg/bootflow/examples/bruteforce/main.go
mimir-d: The bruteforcer is agnostic, yes. But this call still implies knowledge about the flow by virtue of having to know that there is a TPM locality initializer in it (or in a non-trivial example, some other list of measurements). It still seems to me that you have a mirror of the flow in the bruteforce callback func, in order to compute the final hash to compare, whereas in the old design, the Calculate method dealt with this impl aspect, rather than the bruteforcer callback impl (and the Calculate encapsulated the flow logic).
Then again, maybe im misinterpreting something.

pkg/bootflow/examples/ocpdxepcr0/main.go
xaionaro: It was just for syntax sugar.

Talked with @mimir-d in a video chat. Decided to go with `GetChain` instead.

pkg/bootflow/examples/bruteforce/main.go
xaionaro: Talked with @mimir-d: we decided to reuse the TPM Extend logic instead of duplicating it.

pkg/bootflow/bootengine/boot_process.go
xaionaro: Frankly, these details does not seem a priority any soon :)
There are downsides of using an accessor. For example you cannot basically serialize a structure, which does not expose fields. I'm yet not sure how exactly I will handle this.

pkg/bootflow/types/state.go
xaionaro: > Intel ACM would actually have:

I'm actually not sure. It actually might be than `TPMInit(3)` happens only on on-need basis (not earlier).

pkg/bootflow/types/flow.go
```diff
@@ -0,0 +1,29 @@
+package types
+
+// Flow describes steps of the boot process.
+//
+// A flow is static (never change).
+type Flow []Step
+
+// Stop describes a single step of a boot process, essential for the measurements.
```
rihter007, line 8: Stop -> Step?

pkg/bootflow/types/state.go
rihter007: Hm... that looks counter intitive. So different flows should know about each other + that redefinition of the flow for the step.

I would propose to make Flow an interface that returns the next step, this will free flows to know about each other, only share the steps (which they are doing anyway)

```
  type Flow []Step
  
  type Flow interface {
       Next() (Step, error) // returns nil if this is the last step or error is smth went wrong and the next step could not be determined
  }
```

pkg/bootflow/types/trust_chain.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+type TrustChains map[reflect.Type]TrustChain
+
+type TrustChain interface {
```
rihter007, line 11: Could you please provide a couple of examples?
It looks like that we need more methods, to at least check for error

And why do we need IsInitialized?

pkg/bootflow/types/system_artifact.go
```diff
@@ -0,0 +1,28 @@
+package types
+
+import (
+	"fmt"
+	"io"
+	"reflect"
+	"strings"
+)
+
+type SystemArtifacts map[reflect.Type]SystemArtifact
```
rihter007, line 10: Maybe let's use an ID rather then reflect.Type?
This will allow to share code of "firmware part" artifact which all will share same functionality, but just refer to logically different pieces of firmware

pkg/bootflow/types/state.go
```diff
@@ -0,0 +1,125 @@
...
+	key := typeMapKey(systemArtifactSample)
+	systemArtifact := state.SystemArtifacts[key]
+	if systemArtifact == nil {
+		return ErrNoSystemArtifact{SystemArtifactKey: key}
+	}
+
+	return callback(systemArtifact)
+}
+
+func (state *State) EnableSystemArtifact(systemArtifact SystemArtifact) {
```
rihter007, line 79: I would call AddSystemArtifact :)

pkg/bootflow/types/system_artifact.go
rihter007: + what if some steps will share artifacts? (but say have different types of checks or even similar, that can happen in our abstract world)

pkg/bootflow/types/state.go
```diff
@@ -0,0 +1,125 @@
...
+	"reflect"
+	"strings"
+)
+
+// State describes a virtual state of a machine being booted
+type State struct {
+	// Do not mutate these values from the outside, the mutability
+	// is owned by State itself.
+
+	SystemArtifacts SystemArtifacts
```
rihter007, line 14: Will we store somewhere the result of check of an artifact? Is it a TrustChain? Can TrustChain use each other?

rihter007: Ah sorry, seems like SystemArtifact is firmware, while pieces of it that we verify are in VerifiedData?

pkg/bootflow/systemartifacts/biosfirmware/bios_firmware.go
```diff
@@ -0,0 +1,47 @@
...
+		return fn(artifact.(*BIOSFirmware))
+	})
+}
+
+func (fw *BIOSFirmware) Size() uint {
+	return uint(len(fw.Content))
+}
+
+func (fw *BIOSFirmware) ReadAt(b []byte, offset int64) (n int, err error) {
+	return bytes.NewReader(fw.Content).ReadAt(b, offset)
```
rihter007, line 34: Some our interfaces operate on bytes :(

pkg/bootflow/types/trust_chain.go
rihter007: Found examples, sorry :)

pkg/bootflow/systemartifacts/biosfirmware/bios_firmware.go
xaionaro: It will be my job to fix that :)

ChriMarMe:
This got closed due to change of default branch from master to main. Rebase on main and rename branch, open PR again is a legit move here.

pkg/bootflow/examples/ocpdxepcr0/main.go
xaionaro: Changed my mind. Since now we migrated CSS to go1.18, now we can use generics, so in the end I'm going with:
```
tpmInstance := tpm.GetFrom(state)
```
instead of:
```
tpm.StateExec(state, func(tpmInstance *tpm.TPM) error {
```

pkg/bootflow/examples/bruteforce/main.go
xaionaro: Changed to:
```
tpmMeasurements := process.Log.GetDataVerifiedBy(tpmInstance)
```

pkg/bootflow/examples/bruteforce/main.go
xaionaro: Done. Here is the example: https://github.com/9elements/converged-security-suite/blob/a22e7ec0c568acf6c45271250efa9f9ce2827c0e/pkg/bootflow/examples/bruteforce/main.go

pkg/bootflow/actions/commonactions/set_flow.go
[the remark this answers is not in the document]
xaionaro: Fixed by 60c0e0b6aa636851b4a7c0a398d47202383705d8

pkg/bootflow/types/state.go
xaionaro: > Ok, as long as the contract is clear that a flow should be built to leave consistent state behind. 

One of the major idea of this new design: to be close to how actual boot process works (to be able to reproduce its behavior). And unfortunately in the actual boot process we have no such guarantees, thus in these packages not having this guarantees is considered a feature (not a bug): if something in a real boot leaves an unclean state, we want to have exactly the same thing here.

> So different flows should know about each other + that redefinition of the flow for the step.

In reality (in a real boot process) they do know about entry points of each other.

> I would propose to make Flow an interface that returns the next step, this will free flows to know about each other, only share the steps (which they are doing anyway)

Sorry, I do not understand. They will still to know about each other to be able to jump to each other. Otherwise we need to denormalize the pseudo-code tree given above (and make it a bunch of flat sequential steps instead of being a tree).

pkg/bootflow/types/flow.go
```diff
@@ -0,0 +1,29 @@
...
+// Steps of a flow may vary depending on a State.
+//
+// An example: measure specific sections in an AMD Manifest
+type Step interface {
+	Actions(*State) Actions
```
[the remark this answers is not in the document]
xaionaro: There are two semantic layers:
* What to do.
* How to do.

An example of "what to do" is "measure DXE", and corresponding example of "how to do" will be "find UEFI volume, hash it and use TPMExtend and TPMEventLogAdd". Here:
* "Measure DXE" is a `Step`.
* "find UEFI volume [...]" is `Action` (/`Actions`)

For example as you might remember in AMD we have a semantically single step to measure some manifest data, but the implementation requires actually of lot of measurements. So a `Flow` always has constant list of `Step`-s, but a `Step` consts if a list of `Actions` which depends on the State (of the imaginary machine).

pkg/bootflow/types/trust_chain.go
xaionaro: > Could you please provide a couple of examples?

Here is an example of using a `TrustChain`:
https://github.com/9elements/converged-security-suite/blob/60c0e0b6aa636851b4a7c0a398d47202383705d8/pkg/bootflow/examples/ocppeipcr0/main.go

There you can find:
```go
	tpmInstance, err := tpm.GetFrom(state)
	if err != nil {
		panic(err)
	}
```

It extracts `TrustChain` with type `*TPM` from the `state`.

> It looks like that we need more methods, to at least check for error

Unfortunately, different `TrustChain`-s may have different designs (for example DICE and TPM measurements) and because of that we do not have common methods.
But yet we need an interface we will use for the `map` of all `TrustChain`-s.

> And why do we need `IsInitialized`?

That was the only common method I thought of for different `TrustChain`-s. So two reasons:
1. One may use it to know if this `TrustChain` was initialized during a given boot process. If are more common methods for any abstract `TrustChain`, I'd gladly add them here to unlock more opportunities.
2. To do not have `interface{}` and thus to enforce compile time checks that a good type is provided in an argument to somewhere.

pkg/bootflow/types/state.go
xaionaro: The problem with `Add` is when we "add apple, and then again add apple" we get "two apples". While here it is not the case: here "each unique apple can be presented only once".

Renamed to `Include`. Hoping now is better :)

pkg/bootflow/types/state.go
xaionaro: > SystemArtifact is firmware, while pieces of it that we verify are in VerifiedData?

Yes. Though not only "firmware" might be `SystemArtifact`. For example "status registers" is also a `SystemArtifact`.

pkg/bootflow/types/flow.go
xaionaro: Fixed.

PR #335, opened 2022-09-15
feat(pkg/uefi): Add a field to adjust the offset

xaionaro:
Sometimes it is required to do calculations relatively to specific region or volume. But when we parse an UEFI image a component stores the offset inside and we want to avoid reparsing an image (it is expensive). So we add this hack which adjusts offsets without reparsing the image.


ChriMarMe (approved):

u1f35c (approved):
LGTM.

PR #340, opened 2022-11-24
Explain unexpected TPM Event Log entries.

xaionaro:
When looking at TPM EventLog it looks like we have some useful information in there. Here I just automate things I usually do when I try to blindly guess what the measurements (in TPM EventLog) mean.

# Before

```
$ go run ./cmd/pcr0tool/ sum -compare-with-eventlog ~/pcr0tool-cases/case0/tpm_eventlog -registers ~/pcr0tool-cases/case0/registers.json ~/pcr0tool-cases/case0/firmware.fd
[...]
comparing with TPM EventLog result:
	match: false
	updated ACM Policy Status: <nil>
	err: <nil>
	issues: [extra entry in EventLog of type 8 (0x8) on evIdx==1 extra entry in EventLog of type 2147483658 (0x8000000A) on evIdx==2 PCR0_DATA measurement does not match the digest reported in EventLog and unable to brute force a possible bitflip: 62F28D9962FDD86F1B79252C3362DC3B66876202 != 87560D314D6332C2A1521F405C2C00F9599B69F9]
```

# After
```
$ go run ./cmd/pcr0tool/ sum -compare-with-eventlog ~/pcr0tool-cases/case0/tpm_eventlog -registers ~/pcr0tool-cases/case0/registers.json ~/pcr0tool-cases/case0/firmware.fd
[...]
comparing with TPM EventLog result:
	match: false
	updated ACM Policy Status: <nil>
	err: <nil>
	issues:
		* unexpected entry in EventLog of type EV_S_CRTM_VERSION (0x8) on evIdx==1; log entry analysis: reproduced the digest using measurement: {{"ForceData": "0x052B10A7C7D9654181402ADDE94AF63C"}}
		* unexpected entry in EventLog of type EV_EFI_PLATFORM_FIRMWARE_BLOB2 (0x8000000A) on evIdx==2; log entry analysis: reproduced the digest using measurement: {{"Range": {"Offset":"0xac0000", "Length":"0x70000"}}}; related UEFI nodes: [bios_region volume:14E428FA-1A12-4875-B637-8B3CC87FDF07 file:968C1D9F-80C4-43B7-8CAE-668AA56C4E71 file:01FB5D53-4FF9-4AD9-ADF5-7DC63659A1B2 file:BDAD7D1A-4C48-4C75-B5BC-D002D17F6397 file:7ECD9C20-68B9-4A6F-B515-D64FF500B109 file:DACF705C-71DF-497D-AABE-10186B2E1DDE file:9B3F28D5-10A6-46C8-BA72-BD40B847A71A file:961C19BE-D1AC-4BA7-87AF-4AE0F09DF2A6 file:34989D8E-930A-4A95-AB04-2E6CFDFF6631 file:0D8039FF-49E9-4CC9-A806-BB7C31B0BCB0 file:6B844C5B-6B75-42CA-8E8E-1CB94412B59B file:1C8B7F78-1699-40E6-AF33-9B995D16B043 file:89E549B0-7CFE-449D-9BA3-10D8B2312D71 file:39E8CA1A-7A69-4A73-834A-D06381933286 file:D9DDCFF2-215A-480A-AA63-1DF1F5EDBC01 file:8ED25ADD-7C2D-4326-9186-85CCF74C79F6 file:9525E7F7-93CE-4094-B42A-F0DF7E2CF951]
		* unexpected entry in EventLog of type EV_EFI_PLATFORM_FIRMWARE_BLOB2 (0x8000000A) on evIdx==3; log entry analysis: reproduced the digest using measurement: {{"Range": {"Offset":"0x9bf000", "Length":"0x40000"}}}; related UEFI nodes: [bios_region volume:013B9639-D6D5-410F-B7A9-F9173C56ECDA file:0EC4EC45-128B-4B1A-82B5-17BB9344E94B]
		* missing entry (for measurement 'PCR0_DATA') in EventLog (expected event types are: 7 (0x7))
```

Specifically nice to see this stuff:
* `reproduced the digest using measurement: {{"ForceData": "0x052B10A7C7D9654181402ADDE94AF63C"}}`
* `reproduced the digest using measurement: {{"Range": {"Offset":"0xac0000", "Length":"0x70000"}}}`

Indeed:
```
[xaionaro@void typing]$ echo 0x052B10A7C7D9654181402ADDE94AF63C | xxd -r -p | sha1sum
f51bcf65d76500bd86d20b17f5b193fcd4b3b480  -
```
(which matches the Digest in the EventLog entry)

and
```
[xaionaro@void typing]$ dd status=none if=~/pcr0tool-cases/case0/firmware.fd of=/dev/stdout skip=$((16#AC)) count=$((16#7)) bs=$((16#10000)) | sha1sum
5044624aef06c5993bc11d6b6c340600e5c233bf  -
```

This is the entries in from the Log:
```
#                   : 1
PCR index           : 0
Event Type          : 8
Hash Algorithm      : 4
Digest              : F51BCF65D76500BD86D20B17F5B193FCD4B3B480
Data                : ([]uint8) (len=16 cap=16) {
    > 00000000  05 2b 10 a7 c7 d9 65 41  81 40 2a dd e9 4a f6 3c  |.+....eA.@*..J.<|
}
#                   : 2
PCR index           : 0
Event Type          : 2147483658
Hash Algorithm      : 4
Digest              : 5044624AEF06C5993BC11D6B6C340600E5C233BF
Data                : ([]uint8) (len=35 cap=35) {
    > 00000000  12 46 56 5f 42 42 5f 41  46 54 45 52 5f 4d 45 4d  |.FV_BB_AFTER_MEM|
    > 00000010  4f 52 59 00 00 ac ff 00  00 00 00 00 00 07 00 00  |ORY.............|
    > 00000020  00 00 00                                          |...|
}
```

### DIFF
--- a/cmd/pcr0tool/commands/sum/command.go
+++ b/cmd/pcr0tool/commands/sum/command.go
@@ -178,13 +178,13 @@ func (cmd Command) Execute(args []string) {
 	}
 }
 
-func formatIssues(issues []pcrbruteforcer.Issue, intend string) string {
+func formatIssues(issues []pcrbruteforcer.Issue, indent string) string {
 	if len(issues) == 0 {
-		return intend + "NONE"
+		return indent + "NONE"
 	}
 	var result strings.Builder
 	for _, issue := range issues {
-		result.WriteString(fmt.Sprintf("%s* %s\n", intend, issue.Error()))
+		result.WriteString(fmt.Sprintf("%s* %s\n", indent, issue.Error()))
 	}
 	return result.String()
 }
--- a/cmd/pcr0tool/commands/sum/command.go
+++ b/cmd/pcr0tool/commands/sum/command.go
@@ -173,7 +173,18 @@ func (cmd Command) Execute(args []string) {
 		tpmEventLog, err := tpmeventlog.Parse(f)
 		assertNoError(err)
 		match, updatedACMPolicyStatus, issues, err := pcrbruteforcer.ReproduceEventLog(tpmEventLog, tpmeventlog.TPMAlgorithmSHA1, measurements, firmware.Buf(), pcrbruteforcer.DefaultSettingsReproduceEventLog())
-		fmt.Printf("comparing with TPM EventLog result:\n\tmatch: %v\n\tupdated ACM Policy Status: %v\n\terr: %v\n\tissues: %v\n",
-			match, updatedACMPolicyStatus, err, issues)
+		fmt.Printf("comparing with TPM EventLog result:\n\tmatch: %v\n\tupdated ACM Policy Status: %v\n\terr: %v\n\tissues:\n%s\n",
+			match, updatedACMPolicyStatus, err, formatIssues(issues, "\t\t"))
 	}
+}
+
+func formatIssues(issues []pcrbruteforcer.Issue, intend string) string {
+	if len(issues) == 0 {
+		return "NONE"
+	}
+	var result strings.Builder
+	for _, issue := range issues {
+		result.WriteString(fmt.Sprintf("%s* %s\n", intend, issue.Error()))
+	}
+	return result.String()
 }
--- a/cmd/pcr0tool/commands/sum/command.go
+++ b/cmd/pcr0tool/commands/sum/command.go
@@ -180,7 +180,7 @@ func (cmd Command) Execute(args []string) {
 
 func formatIssues(issues []pcrbruteforcer.Issue, intend string) string {
 	if len(issues) == 0 {
-		return "NONE"
+		return intend + "NONE"
 	}
 	var result strings.Builder
 	for _, issue := range issues {
--- a/pkg/pcrbruteforcer/analyze_unexpected_log_entry.go
+++ b/pkg/pcrbruteforcer/analyze_unexpected_log_entry.go
@@ -1,0 +1,211 @@
+package pcrbruteforcer
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/9elements/converged-security-suite/v2/pkg/diff"
+	"github.com/9elements/converged-security-suite/v2/pkg/pcr"
+	"github.com/9elements/converged-security-suite/v2/pkg/tpmeventlog"
+	"github.com/9elements/converged-security-suite/v2/pkg/uefi"
+	"github.com/9elements/converged-security-suite/v2/pkg/uefi/ffs"
+	pkgbytes "github.com/linuxboot/fiano/pkg/bytes"
+)
+
+const (
+	PhysAddrBase = tpmeventlog.PhysAddrBase
+)
+
+type logEntryExplanation struct {
+	Measurement     *pcr.Measurement
+	RelatedNodes    []diff.NodeInfo
+	EventDataParsed *tpmeventlog.EventDataParsed
+}
+
+func formatMeasurement(m pcr.Measurement) string {
+	var result []string
+	for _, chunk := range m.Data {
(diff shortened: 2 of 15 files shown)


cmd/pcr0tool/commands/sum/command.go
```diff
@@ -173,7 +173,18 @@ func (cmd Command) Execute(args []string) {
...
-		fmt.Printf("comparing with TPM EventLog result:\n\tmatch: %v\n\tupdated ACM Policy Status: %v\n\terr: %v\n\tissues: %v\n",
-			match, updatedACMPolicyStatus, err, issues)
+		fmt.Printf("comparing with TPM EventLog result:\n\tmatch: %v\n\tupdated ACM Policy Status: %v\n\terr: %v\n\tissues:\n%s\n",
+			match, updatedACMPolicyStatus, err, formatIssues(issues, "\t\t"))
 	}
 }
+
+func formatIssues(issues []pcrbruteforcer.Issue, intend string) string {
+	if len(issues) == 0 {
+		return "NONE"
```
u1f35c, line 183: Should this be prefixed with intend?

pkg/pcrbruteforcer/analyze_unexpected_log_entry.go
```diff
@@ -0,0 +1,211 @@
...
+	if expectedMeasurement != nil {
+		measurementID = expectedMeasurement.ID
+	}
+	measurement := &pcr.Measurement{
+		ID:   measurementID,
+		Data: chunks,
+	}
+	h, err := ev.Digest.HashAlgo.Hash()
+	if err != nil {
+		panic(err) // should never happen
```
u1f35c, line 134: Can we avoid adding panics to the code? Even if they should never happen they turn up and kill the service.

xaionaro: Considering it should never happen it seems unreasonable to make the signature of the function more complex and require to process errors (which will never happen) upper-side. But this is not a strong opinion, so feel free to insist ( then will fix :) )

cmd/pcr0tool/commands/sum/command.go
xaionaro: Good catch.

pkg/pcrbruteforcer/analyze_unexpected_log_entry.go
u1f35c: If it's never going to happen then can we not either ignore the error, or just return nil?

Given it looks like this could only fail if we fail to create a Hash object or calculate a digest (below) this shouldn't be affected by untrusted BIOS image input, and really shouldn't fail, so I'll not push hard for the removal of the panic.

xaionaro: > If it's never going to happen then can we not either ignore the error, or just return nil?

Personally I hate silent errors even more than panics :)
This error should never happen by design. But since an error value is returned by this function and I have choice to:
* Process it everywhere expecting to never happen (and thus complicate code without reason).
* Ignore it.
* Panic.

My priorities are:
panic > process > ignore

If I'm wrong about "should never happen by design" then better to surface the error than silently ignore it. In future we should be able to add go-belt observability into the application and replace this `panic(err)` with harmless `errmon.ObserveError(err)`.

cmd/pcr0tool/commands/sum/command.go
xaionaro: Fixed.

u1f35c (approved):

zaolin (approved):

PR #344, opened 2022-12-02
Fix PCR OCP PCD Firmware Version

xaionaro:
The mechanism to return the default PCD Firmware Version String is broken. Fixing it.


pkg/pcr/measurement_id.go
```diff
@@ -399,7 +399,7 @@ func (id MeasurementID) MeasureFunc() MeasureFunc {
 	}
 	return func(config MeasurementConfig, provider DataProvider) (Measurements, error) {
 		m, err := measureFunc(config, provider)
-		if err != nil {
+		if m == nil {
 			return nil, err
```
xaionaro, line 403: The fix itself.

When we use the default value, instead of parsing, we also return that warning as `err`.

zaolin (approved):

PR #347, opened 2023-01-23
feat(pcr0tool): Add support of a logger

xaionaro:
Adding basic support of a Logger to `pcr0tool`.

# Test Plan

Ran:
```
go run ./cmd/pcr0tool/ -log-level=trace sum -hash-func sha256 -registers ~/firmware/northdome.registers ~/firmware/northdome.fd
```
Output:
```
DEBU[0000] FindEmbeddedFirmwareStructure err-result: <nil>
debugInfo: {
  "config_orig": {
    "Flow": 0,
    "FindMissingFakeMeasurements": false,
    "Registers": [
      {
        "id": "MP0_C2P_MSG_37",
        "value": "jRAAEQ=="
      },
      {
        "id": "MP0_C2P_MSG_38",
        "value": "AAAAUA=="
      }
    ],
    "PCR0DataIbbDigestHashAlgorithm": 11,
    "TPMDevice": 0
  },
  "config_result": {
    "Flow": 6,
    "FindMissingFakeMeasurements": false,
    "Registers": [
      {
        "id": "MP0_C2P_MSG_37",
        "value": "jRAAEQ=="
      },
      {
        "id": "MP0_C2P_MSG_38",
        "value": "AAAAUA=="
      }
    ],
    "PCR0DataIbbDigestHashAlgorithm": 11,
    "TPMDevice": 0
  },
  "detectedAttestationFlow": "LegacyAMDLocality3",
  "warnings": "errors: unable to collect measurement 'pcdFirmwareVendor_measured_data' (is_fake:false): unable to find the source of firmware vendor version"
}
measurements: [
  {
    "ID": "PSP firmware version",
    "Data": [
      {
        "Range": {
          "Offset": 1778784,
          "Length": 4
        }
      }
    ]
  },
  {
    "ID": "BIOS RTM Volume",
    "Data": [
      {
        "Range": {
          "Offset": 13041664,
          "Length": 3735552
        }
      }
    ]
  },
  {
    "ID": "pcdFirmwareVendor_measured_data",
    "Data": [
      {
        "Range": {
          "Offset": 0,
          "Length": 0
        },
        "ForceData": "HvtrVAwdVUCkrU70vxe4Og=="
      }
    ]
  },
  {
    "ID": "DXE",
    "Data": [
      {
        "Range": {
          "Offset": 5644288,
          "Length": 7397376
        }
      }
    ]
  },
  {
    "ID": "separator",
    "Data": [
      {
        "Range": {
          "Offset": 0,
          "Length": 0
        },
        "ForceData": "AAAAAA=="
      }
    ]
  }
]
Set 0x -> 0x0000000000000000000000000000000000000000000000000000000000000003

Event 'PSP firmware version': 6e001300 (*sha256.digest)
*sha256.digest(0x 0000000000000000000000000000000000000000000000000000000000000003 625C8E327941BDFB9099E21223DA6A76506A53F78150731CC15A46C5A69B5F49) == 0xA64E77E3E58BE712166AAD6094EDADB54835C700C287BF824FEDDC211F058C93

Event 'BIOS RTM Volume': 0000000000000000000000000000000078e58c8c... (len: 3735552) (*sha256.digest)
*sha256.digest(0x A64E77E3E58BE712166AAD6094EDADB54835C700C287BF824FEDDC211F058C93 B03975AC0D62E52C907CA1B219FB17847F54D33F49088C45BBEC19FF7662B8B7) == 0x82AA886E4FCC39440AFF481E743FF3D7F1EC8A254264147AC554B40F7243901C

Event 'pcdFirmwareVendor_measured_data': 1efb6b540c1d5540a4ad4ef4bf17b83a (*sha256.digest)
*sha256.digest(0x 82AA886E4FCC39440AFF481E743FF3D7F1EC8A254264147AC554B40F7243901C D4720B4009438213B803568017F903093F6BEA8AB47D283DB32B6EABEDBBF155) == 0x69422284A65D8494AF6A2F1EDC597759DE77B055EDD59E115D807A9B86D04E96

Event 'DXE': 000000000000000000000000000000007ac07354... (len: 7397376) (*sha256.digest)
*sha256.digest(0x 69422284A65D8494AF6A2F1EDC597759DE77B055EDD59E115D807A9B86D04E96 327885A92725B24F36664245B79517F7C2E1EB8DEDE90489540508250821A430) == 0x12E7B33F7160E7AFDE19A1D8A1BB3CC7F42615228B1FA706DAD88A2FCE701393

Event 'separator': 00000000 (*sha256.digest)
*sha256.digest(0x 12E7B33F7160E7AFDE19A1D8A1BB3CC7F42615228B1FA706DAD88A2FCE701393 DF3F619804A92FDB4057192DC43DD748EA778ADC52BC498CE80524C014B81119) == 0x64C87B9C60B73B405A8E21323D7BFFB67A8FF39D8B095DF1FEBE2988FF9CA4CC

Resulting PCR0: 64C87B9C60B73B405A8E21323D7BFFB67A8FF39D8B095DF1FEBE2988FF9CA4CC
```

The first line is:
```
DEBU[0000] FindEmbeddedFirmwareStructure err-result: <nil>
```
(as expected)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/digitalocean/go-smbios v0.0.0-20180907143718-390a4f403a8e
 	github.com/edsrzf/mmap-go v1.1.0
-	github.com/facebookincubator/go-belt v0.0.0-20230703220829-b6b46c95ec1f
+	github.com/facebookincubator/go-belt v0.0.0-20230703220935-39cd348f1a38
 	github.com/fearful-symmetry/gomsr v0.0.1
 	github.com/go-ng/slices v0.0.0-20230703171042-6195d35636a2
 	github.com/go-ng/xmath v0.0.0-20230704233441-028f5ea62335
--- a/go.sum
+++ b/go.sum
@@ -111,8 +111,8 @@ github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.m
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
 github.com/envoyproxy/protoc-gen-validate v0.0.14/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
-github.com/facebookincubator/go-belt v0.0.0-20230703220829-b6b46c95ec1f h1:FW/+mUF6GAsdTz/ulSMpcBE204rnOTrPmMyd4d7pPDE=
-github.com/facebookincubator/go-belt v0.0.0-20230703220829-b6b46c95ec1f/go.mod h1:GX1P3GiO+6E2SxTZaI3z1cqlTGfE1GnKxEsxBVemVnY=
+github.com/facebookincubator/go-belt v0.0.0-20230703220935-39cd348f1a38 h1:6bNfYFYry8V+BHinBMDEPPnd19unWa76lQhsx88xU6k=
+github.com/facebookincubator/go-belt v0.0.0-20230703220935-39cd348f1a38/go.mod h1:GX1P3GiO+6E2SxTZaI3z1cqlTGfE1GnKxEsxBVemVnY=
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
 github.com/fearful-symmetry/gomsr v0.0.1 h1:m208RzdTApWVbv8a9kf78rdPLQe+BY9AxRb/nSbHxSA=
 github.com/fearful-symmetry/gomsr v0.0.1/go.mod h1:Qb/0Y7zwobP7v8Sji+M5mlL4N7Voyz5WaKXXRFPnLio=


pkg/log/fiano_logger.go
```diff
@@ -0,0 +1,32 @@
+package log
+
+import (
+	"github.com/facebookincubator/go-belt/tool/logger"
+	fianoLog "github.com/linuxboot/fiano/pkg/log"
+)
+
+var _ fianoLog.Logger = FianoLogger{}
+
+// FianoLogger is just a placeholder for a logger, which does nothing
```
u1f35c, line 10: This isn't true? We've gone from it doing nothing to being a wrapper around a logger. Also should the methods below not be checking if Backend is nil?

pkg/log/fiano_logger.go
```diff
@@ -0,0 +1,32 @@
...
+type FianoLogger struct {
+	Backend    logger.Logger
+	LogAsLevel logger.Level
```
u1f35c, line 13: Do these need to be exported? I don't see them used outside this package?

xaionaro: Any reason to hide it? :)

pkg/log/fiano_logger.go
xaionaro: Yeah, replaced `DummyLogger` with `FianoLogger`, but forgot to update the comments. Fixed.

pkg/log/fiano_logger.go
u1f35c: I generally lean towards not exporting things that aren't needed so interface boundaries are kept to only the required pieces, but no strong push towards that here given it's all internal.

pkg/log/fiano_logger.go
u1f35c: What about the nil checks? You do `fianoLog.DefaultLogger = log.FianoLogger{}` in various places, which I think means if Fiano tries to log it's going to try to call a function on Backend which will be nil?

pkg/log/fiano_logger.go
xaionaro: It seems to me C++-y approach. I'd prefer in Go to export by default if there are no explicit safety concerns. In my opinion in Go we need to protect fields only if there are hidden assumptions about these fields.

pkg/log/fiano_logger.go
xaionaro: IDE replaced it everywhere (I used this function) and I just didn't know about those cases :(

Thanks for noticing! Fixed.

pkg/log/fiano_logger.go
u1f35c: Not insisting here, but let's have a conversation about it when I'm back after PTO.

u1f35c (approved):

ChriMarMe:
I know this took way too long on my side, sorry for that.

Can you rebase and fix, push and i will happily review and merge this week.

ChriMarMe (approved):
LGTM. Can be merged when rebased :)

xaionaro:
The PR is beyond recovery by now.

PR #349, opened 2023-01-23
Adjustments for a new AMD platform

xaionaro:
# Please review on per commit basis.

---

# Additional context on commit https://github.com/9elements/converged-security-suite/pull/349/commits/af5cf6d67867864834abdc09f069b56de50fd1b6 :

Basically I want to get
`* measurement 'DXE' does not match the digest reported in EventLog: 327885A92725B24F36664245B79517F7C2E1EB8DEDE90489540508250821A430 != FB4DA84CADAB0FF6ABD9AB6354D850A472C2F245D6EFA8DB6D4B9FE47D525EE3; log entry analysis: mentioned byte ranges: [{"Offset":"0xff5630a4", "Length":"0x70cf5c"}]; related UEFI nodes: [bios_region volume:4F1C52D3-D824-4D2A-A2F0-EC40C23C5916 file:9E21FD93-9C72-4C15-8C4B-E77F1DB2D792]; possible digest: 500E22B42E2732F6CB7B626D3491F3650440BBC47390B5A07CB2168D70F42078`

instead of  just:
`* measurement 'DXE' does not match the digest reported in EventLog: 327885A92725B24F36664245B79517F7C2E1EB8DEDE90489540508250821A430 != FB4DA84CADAB0FF6ABD9AB6354D850A472C2F245D6EFA8DB6D4B9FE47D525EE3; log entry analysis: mentioned byte ranges: [{"Offset":"0xff5630a4", "Length":"0x70cf5c"}]; related UEFI nodes: [bios_region volume:4F1C52D3-D824-4D2A-A2F0-EC40C23C5916 file:9E21FD93-9C72-4C15-8C4B-E77F1DB2D792]`

The additional digest also provides information. So what I have in my case:
* We expect one digest.
* In TPM EventLog we have another digest.
* The referenced data (in the description of the log entry) has a third digest.

In this case I cannot reproduce PCR0 from TPM EventLog, so this third digest is an useful hint which exactly digest there could be instead.


zaolin (approved):

pkg/pcrbruteforcer/reproduce_event_log.go
```diff
@@ -20,12 +20,14 @@ import (
...
 type SettingsReproduceEventLog struct {
 	SettingsBruteforceACMPolicyStatus
+	DisabledEventsMaxDistance uint64
 }
 
 // DefaultSettingsReproduceEventLog returns recommended default PCR0 settings
 func DefaultSettingsReproduceEventLog() SettingsReproduceEventLog {
 	return SettingsReproduceEventLog{
 		SettingsBruteforceACMPolicyStatus: DefaultSettingsBruteforceACMPolicyStatus(),
+		DisabledEventsMaxDistance:         2, // arbitrary value based on previous experience, hoping to handle within a second; TODO: add benchmarks
```
u1f35c, line 30: We drop from 5 to 2 without a comment about the fact we're doing so in the commit message.

u1f35c (approved):

pkg/pcrbruteforcer/reproduce_event_log.go
xaionaro: Yeah. AMD has so many measurements that bruteforcing becomes too expensive computationally. Because of that I had to drop the depth of bruteforcing.

PR #359, opened 2023-04-13
[pkg/tpmeventlog] Move Replay from `pcr` to `tpmeventlog`

xaionaro:
This is required to migrate from `pcr` to `bootflow`.


pkg/tpmeventlog/replay.go
```diff
@@ -50,3 +52,89 @@ func (eventLog *TPMEventLog) FilterEvents(pcrIndex pcr.ID, hashAlgo TPMAlgorithm
...
+				return nil, fmt.Errorf("unable to hash: %w", err)
+			}
+			result = hasher.Sum(nil)
+			hasher.Reset()
+			_, _ = fmt.Fprintf(logOut, " -> %X\n", result)
+		}
+	}
+
+	return result, nil
+}
```
xaionaro, line 140: Moved from pcr/replay.go, with few changes:
* Deleted the first `case` in the `switch`.
* Updated the condition in the second `case`.
* Removed `tpmeventlog.`.

pkg/tpmeventlog/errors.go
```diff
@@ -2,6 +2,8 @@ package tpmeventlog
 
 import (
 	"fmt"
+
+	pcr "github.com/9elements/converged-security-suite/v2/pkg/pcr/types"
```
xaionaro, line 6: This is a tolerable (and lightweight enough) dependency for now.

u1f35c (approved):

PR #360, opened 2023-04-18
[pcr] Fix panic in pcr.ValidateManifests.Validate

xaionaro:
Otherwise a panic occurs in `github.com/9elements/converged-security-suite/v2/pkg/pcr/validate_manifests.go:29`.

### DIFF
--- a/pkg/pcr/get_measurements_pcr0_cbnt0t.go
+++ b/pkg/pcr/get_measurements_pcr0_cbnt0t.go
@@ -161,7 +161,11 @@ func getBootPolicyManifest(fitEntries []fit.Entry) (*cbntbootpolicy.Manifest, *f
 		switch fitEntry := fitEntry.(type) {
 		case *fit.EntryBootPolicyManifestRecord:
 			_, bpManifest2, err := fitEntry.ParseData()
-			if err != nil && bpManifest2 != nil {
+			if bpManifest2 == nil {
+				// TODO: add support for non-CBnT BtG
+				return nil, nil, fmt.Errorf("found a non-CBnT BootGuard manifest instead of the CBnT one")
+			}
+			if err != nil {
 				return nil, nil, err
 			}
 			return bpManifest2, fitEntry, nil


pkg/pcr/flow_detection.go
```diff
@@ -243,15 +243,24 @@ func isCBnT(fitEntries []fit.Entry) (bool, error) {
 	for _, fitEntry := range fitEntries {
 		switch fitEntry := fitEntry.(type) {
 		case *fit.EntryKeyManifestRecord:
-			data, data2, err := fitEntry.ParseData()
-			if data == nil && data2 == nil {
+			_, data2, err := fitEntry.ParseData()
```
u1f35c, line 246: Can we just have "data" instead of "data2" now there's only a single data variable? Likewise below.

pkg/pcr/get_measurements_pcr0_cbnt0t.go
```diff
@@ -71,33 +73,63 @@ func MeasurePCR0Data(config MeasurementConfig, imageSize uint64, fitEntries []fi
 		Length: uint64(len(acmEntry.GetRSASig())),
 	}
 
-	keyManifest, keyManifestFITEntry, err := getKeyManifest(fitEntries)
+	keyManifest1, keyManifest2, keyManifestFITEntry, err := getKeyManifest(fitEntries)
```
u1f35c, line 76: These are different types of manifests and it would be good if the variables they are assigned to could reflect this rather than just having a 1/2 suffix (same comment applies to the later calls to `getKeyManifest`)

u1f35c (changes requested):

pkg/pcr/flow_detection.go
xaionaro: I changed the naming, but not exactly as requested, so let me know if still needs improving :)

pkg/pcr/get_measurements_pcr0_cbnt0t.go
xaionaro: Change a naming a bit (and added comments), let me know if not good enough.

u1f35c (approved):

PR #364, opened 2023-07-06
feat(pcr0tool): Add verb `pcrread`

xaionaro:
Just a simple tool to read a PCR value from TPM (without using `tpm2-tools`).

Feel free to reject this PR. I only added this PR for a single one-time investigation to differentiate if the bug is in tpm2-tools or with TPM module (or its kernel driver) itself.


ChriMarMe:
What is the status here @rihter007 @trynity @u1f35c ?

cmd/pcr0tool/main.go
```diff
@@ -33,6 +34,7 @@ var knownCommands = map[string]commands.Command{
 	"dump_registers":               &dumpregisters.Command{},
 	"printnodes":                   &printnodes.Command{},
 	"validate_security":            &validatesecurity.Command{},
+	"pcrread":                      &pcrread.Command{},
```
trynity, line 37: nit: Please sort these commands alphabetically

trynity (approved):
One minor quibble on the `panic` and a nit, but approving to unblock.

cmd/pcr0tool/main.go
xaionaro: Fixed.

cmd/pcr0tool/commands/pcrread/command.go
```diff
@@ -0,0 +1,79 @@
...
+	assertNoError(err)
+
+	hashAlgo := tpm2.AlgUnknown
+	for _, alg := range []tpm2.Algorithm{tpm2.AlgSHA1, tpm2.AlgSHA256} {
+		if strings.EqualFold(*cmd.hashAlgo, alg.String()) {
+			hashAlgo = alg
+		}
+	}
+	if hashAlgo == tpm2.AlgUnknown {
+		panic(fmt.Errorf("algo '%s' is unknown", *cmd.hashAlgo))
```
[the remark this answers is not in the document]
xaionaro: If the tool is executed with invalid arguments, then the execution should be interrupted. Replaced with `log.Fatal` to be consistent with the rest of the code. Let me know if you'd like another solution.

PR #373, opened 2024-01-23
Feat/BTGimprovements

walterchris:


pkg/provisioning/bootguard/bootguard.go
```diff
@@ -888,20 +888,41 @@ func (b *BootGuard) BPMKeyMatchKMHash() (bool, error) {
...
 
-// SaneBPMSecurityProps verifies that BPM contains security properties set accordingly to spec
-func (b *BootGuard) SaneBPMSecurityProps() (bool, error) {
+func (b *BootGuard) StrictSaneBPMSecurityProps() (bool, error) {
 	switch b.Version {
 	case bgheader.Version10:
 		flags := b.VData.BGbpm.SE[0].Flags
-		if !flags.DMAProtection() {
-			return false, fmt.Errorf("dma protection should be enabled for bootguard")
-		}
```
ChriMarMe, line 898: If this is strict, why remove the check here?

pkg/provisioning/bootguard/bootguard.go
```diff
@@ -888,20 +888,41 @@ func (b *BootGuard) BPMKeyMatchKMHash() (bool, error) {
...
+}
+
+// SaneBPMSecurityProps verifies that BPM contains security properties set accordingly to spec
+func (b *BootGuard) SaneBPMSecurityProps() (bool, error) {
+	switch b.Version {
+	case bgheader.Version10:
+		flags := b.VData.BGbpm.SE[0].Flags
+		if !flags.DMAProtection() {
+			return false, fmt.Errorf("dma protection should be enabled for bootguard")
+		}
```
ChriMarMe, line 925: If this is not strict: Why is this function here then?

pkg/provisioning/bootguard/bootguard.go
```diff
@@ -888,20 +888,41 @@ func (b *BootGuard) BPMKeyMatchKMHash() (bool, error) {
 	return true, nil
 }
 
-// SaneBPMSecurityProps verifies that BPM contains security properties set accordingly to spec
```
ChriMarMe, line 891: Instead of removing the comment, it would be better to adapt it :)

pkg/provisioning/bootguard/bootguard.go
walterchris: This is not strict. It just how git/github renders this for you. Or wdy mean?

pkg/provisioning/bootguard/bootguard.go
walterchris: See above.

pkg/provisioning/bootguard/bootguard.go
ChriMarMe: The function name is `StrictSaneBPMSecurityProps`, but for me it seems the check `!flags.DMAProtection()` is removed from it and placed in the function `SaneBPMSecurityProps`. I would assume this check should be kept in the the function `StrictSaneBPMSecurityProps`, because the function name suggests that it is the more `strict` variant of `SaneBPMSecurityProps`.

walterchris: No. There was only one function SaneBPMSecurityProps where all checks where in, I added StrictSaneBPMSecurityProps and sorted the checks accordingly.

It's just GitHub showing it a bit weird for you.

ChriMarMe: So the check just belongs to the `sane`-function and not to `strictsane`-function :)

ChriMarMe (approved):

PR #374, opened 2024-01-30
Update go version to 1.21

ChriMarMe:


ChriMarMe:
Here we go. Now all circleCI jobs work on the same go version (1.21)

PR #380, opened 2024-02-14
feat: migrate to Github Actions

walterchris:
Migrate our current CI to github actions. Additionally remove the .rpm and .deb builds on version tags. Add template generation for bg-prov.


ChriMarMe (approved):

walterchris:
> Dunno why still blocked. Complains about not signed commit, but commit is signed?!

Fixed. I enabled "signed commits" and _thought_ it would mean that a commit needs a Signed-off line, however it means signing the commit with a valid SSH or GPG Key.

PR #388, opened 2024-03-23
feat: restructure cmd folder

walterchris:
Dividing the cmd folder structure into exp and core commands. Commands that are within the exp folder are not tested 'at all' - and we most likely do not have the infrastructure to test it. Commands in the core folder have a better chance to get tested.


walterchris:
Merging this without review.

PR #423, opened 2026-03-27
ci: fix liner errors

micgor32:
Currently liner check fails with:
```
35 issues:
  * errcheck: 10
  * staticcheck: 25
```
Note: the CI does not check `cmd/`, I covered the errcheck errors that can be seen by running the linter locally over `cmd/`, all staticcheck error are not covered yet. I think it would make sense to modify the workflow to also go over `cmd/`, but that's something for the future (as it requires quite a lot of fixing for the linter to be "green")

### DIFF
--- a/cmd/core/bg-prov/cmd.go
+++ b/cmd/core/bg-prov/cmd.go
@@ -313,7 +313,10 @@ func (kmp *kmPrintCmd) Run(ctx *context) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
+	defer func() {
+		err := file.Close()
+		log.Warnf("failed to close the file: %v\n", err)
+	}()
 	bg, err := bootguard.NewKM(file)
 	if err != nil {
 		return err
@@ -327,7 +330,10 @@ func (bpmp *bpmPrintCmd) Run(ctx *context) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
+	defer func() {
+		err := file.Close()
+		log.Warnf("failed to close the file: %v\n", err)
+	}()
 	bg, err := bootguard.NewBPM(file)
 	if err != nil {
 		return err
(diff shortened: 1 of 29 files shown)


cmd/core/bg-prov/cmd.go
```diff
@@ -327,7 +330,10 @@ func (bpmp *bpmPrintCmd) Run(ctx *context) error {
 	if err != nil {
 		return err
 	}
-	defer file.Close()
+	defer func() {
+		err := file.Close()
+		log.Warnf("failed to close the file: %v\n", err)
```
walterchris: What if there was no error closing the file? Doesn't it miss an if err != nil ?!

cmd/exp/pcr0tool/commands/bruteforce_acm_policy_status/command.go
```diff
@@ -190,7 +190,10 @@ func (cmd Command) Execute(ctx context.Context, args []string) {
 			ctx.tpmCommands[commandIdx].(*tpm.CommandExtend).Digest = newDigest[:]
 
 			ctx.tpm.Reset()
-			ctx.tpm.TPMExecute(context.Background(), ctx.tpmCommands, nil)
+			err := ctx.tpm.TPMExecute(context.Background(), ctx.tpmCommands, nil)
+			if err != nil {
+				panic(err)
```
walterchris, line 195: Why are we panic here?

walterchris (changes requested):
The first comments is valid for 60% of the changes.

cmd/exp/pcr0tool/commands/bruteforce_acm_policy_status/command.go
micgor32: `Execute(ctx context.Context, args []string)` panics on every error.

cmd/core/bg-prov/cmd.go
micgor32: done

walterchris (approved):

PR #424, opened 2026-04-07
CBnT 2.1 and adapting to Fiano changes

micgor32:
Add support for CBnT 2.1 (MTL and newer generations) and adapts to Fiano `intel/metadata` breaking changes.
Short summary of changes:
 - use unified `bg` and `cbnt` packages: also implies that KM and BPM are no longer versioned. Because of the unification, the constructors for the Manifests do the job of returning the correct implementation of either KM or BPM based on the BtG/CBnT version provided. This also implies that we have to make type assertions in few places.
 - add two new tests that are CBnT 2.1 specific to bg-suite. The rationale is given in `cmd/core/bg-suite/TESTPLAN.md`. Because some runtime tests (5, 6 and 8) are only valid on either 1.0/2.0 or 2.1, add the logic for determining the BtG/CBnT version on runtme.
 - read sysfs to get ME version. Current HFSTS definitions are no longer valid for MTL and newer plaftorms that use ME 18.x/19.x/21.x.
 - BPM test (test 3) throws warnings instead of a hard fail when running with strict flag.


micgor32:
force pushed to amend previous commit message :D

micgor32:
squashed fixup commits

micgor32:
depends on https://github.com/linuxboot/fiano/pull/450

ChriMarMe (approved):

PR #425, opened 2026-05-04
ci: add round trip validation job

micgor32:
Add CI job that validates that the binaries exported from a given image are identical to the ones generated based on the config from the same image.

When calculating IBB segments, let the bios flash area size to be taken from IFD on a missmatch between the image length and the bios flash area.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,22 +140,33 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Checkout repository
-              uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+              uses: actions/checkout@v6.0.1
 
             - name: Install dependencies
               run: sudo apt update & sudo apt install openssl
 
+            - name: Cache firmware images
+              id: cache-firmware
+              uses: actions/cache@v4
+              with:
+                path: |
+                  firmware10.bin
+                  firmware20.bin
+                key: firmware-imgs
+
             - name: Download FW with BtG 1.0
+              if: steps.cache-firmware.outputs.cache-hit != 'true'
               run: |
                 wget "https://downloads.hpe.com/pub/softlib2/software1/pubfw-uefi/p736852486/v283550/U30_3.66_04_01_2026.signed.flash"
                 mv U30_3.66_04_01_2026.signed.flash firmware10.bin
             - name: Download FW with CBnT 2.0
+              if: steps.cache-firmware.outputs.cache-hit != 'true'
               run: |
                 wget "https://download.asrock.com/BIOS/4677/W790%20WS(9.01)ROM.zip"
                 unzip W790\ WS\(9.01\)ROM.zip
                 mv W790-WS_9.01.ROM firmware20.bin
             - name: Download Artifacts
-              uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+              uses: actions/download-artifact@v7.0.0
               with:
                 name: artifacts-amd64
                 path: ./artifacts
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,3 +135,97 @@ jobs:
                 --pbet=12 --ibbflags=1 --mchbar=123456 --vdtbar=120000 --dmabase0=130000 \
                 --dmasize0=2048 --entrypoint=140000 --ibbhash=SHA256 config.json
                 cat ./config.json | jq
+    RoundTripValidation:
+        needs: build
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+            - name: Install dependencies
+              run: sudo apt update & sudo apt install openssl
+
+            - name: Download FW with BtG 1.0
+              run: |
+                wget "https://downloads.hpe.com/pub/softlib2/software1/pubfw-uefi/p736852486/v283550/U30_3.66_04_01_2026.signed.flash"
+                mv U30_3.66_04_01_2026.signed.flash firmware10.bin
+            - name: Download FW with CBnT 2.0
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -135,3 +135,97 @@ jobs:
                 --pbet=12 --ibbflags=1 --mchbar=123456 --vdtbar=120000 --dmabase0=130000 \
                 --dmasize0=2048 --entrypoint=140000 --ibbhash=SHA256 config.json
                 cat ./config.json | jq
+    RoundTripValidation:
+        needs: build
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```
walterchris: Why are you not pinning a specific version instead of a commit hash?

.github/workflows/build.yml
```diff
@@ -135,3 +135,97 @@ jobs:
...
+
+            - name: Download FW with BtG 1.0
+              run: |
+                wget "https://downloads.hpe.com/pub/softlib2/software1/pubfw-uefi/p736852486/v283550/U30_3.66_04_01_2026.signed.flash"
+                mv U30_3.66_04_01_2026.signed.flash firmware10.bin
+            - name: Download FW with CBnT 2.0
+              run: |
+                wget "https://download.asrock.com/BIOS/4677/W790%20WS(9.01)ROM.zip"
+                unzip W790\ WS\(9.01\)ROM.zip
+                mv W790-WS_9.01.ROM firmware20.bin
```
walterchris, line 167: Can we cache the above? At least the FW downloads?

This will break once the files are not available anymore.

.github/workflows/build.yml
```diff
@@ -135,3 +135,97 @@ jobs:
...
+            - name: Download Artifacts
+              uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
```
walterchris: Same - Why commit hash and not version?

.github/workflows/build.yml
```diff
@@ -135,3 +135,97 @@ jobs:
...
+                ./artifacts/bg-prov km-export ./firmware10.bin km10.bin
+                sha256sum km10.bin > checksums
+                sha256sum bpm10.bin >> checksums
+                sha256sum km10.bin
+                sha256sum bpm10.bin
+                rm km10.bin bpm10.bin
+            - name: Extract pubkey from KM 1.0
+              run: |
+                KEY_DATA_B64=$(jq -r '."v1-keymanifest".kmKeySignature.ksKey.keyData' config10.json)
+                echo "$KEY_DATA_B64" | base64 -d > /tmp/keydata.bin
```
walterchris: Wouldn't `jq -r '."v1-keymanifest".kmKeySignature.ksKey.keyData' config10.json | base64 -d > ...` also work?

.github/workflows/build.yml
```diff
@@ -135,3 +135,97 @@ jobs:
...
+                sha256sum -c checksums
+            - name: Read 2.0 config
+              run: |
+                ./artifacts/bg-prov read-config config20.json ./firmware20.bin
+                ./artifacts/bg-prov bpm-export ./firmware20.bin bpm20.bin
+                ./artifacts/bg-prov km-export ./firmware20.bin km20.bin
+                sha256sum km20.bin > checksums
+                sha256sum bpm20.bin >> checksums
+                rm km20.bin bpm20.bin
+            - name: Extract pubkey from KM 2.0
```
walterchris, line 204: Any chance we can move this into it's own action to avoid deduplication?

.github/workflows/build.yml
```diff
@@ -135,3 +135,97 @@ jobs:
...
+                  -inform DER \
+                  -RSAPublicKey_in \
+                  -pubout \
+                  -out km_pub20.pem
+                rm -f /tmp/keydata.bin /tmp/rsa_key.asn1 /tmp/rsa_key.der
+            - name: Generate 2.0 from config
+              run: |
+                ./artifacts/bg-prov bpm-gen-v-2 --config=config20.json bpm20.bin ./firmware20.bin
+                ./artifacts/bg-prov km-gen-v-2 --config=config20.json km20.bin km_pub20.pem
+                sha256sum -c checksums
```
walterchris, line 211: Help me.. why does this work?

walterchris (changes requested):

.github/workflows/build.yml
micgor32: sure, there is only a small uncommon part here - the `v1-keymanifest".kmKeySignature.ksKey.keyData` and `v2-keymanifest".kmKeySignature.ksKey.keyData`, so no problem in removing the duplication

.github/workflows/build.yml
micgor32: we write the checksums of the exported BPM and KM to a file, then remove the exported bins, and generate the new ones based on the config. With -c flag on sha256sum it takes input of the file and does the check on all matching files under the current directory [1]. So the checksums file looks like this:
```
ee623447ed8d6c39d6466e641cb33e9c32b3d4ad1df9d450897e61014e2cbdbc  bpm20.bin
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa  km20.bin
```
So if we name the new generated ones the same and run sha256sum -c, it passes only if these are matching.

[1] - https://www.man7.org/linux/man-pages/man1/sha256sum.1.html

micgor32: the whole magic for extracting the pubkeys is there to have exactly the same pubkey in the generated KM. It is there in the json config that we read, but its base64 encoded and the generator requires valid rsa pubkey in .pem file.

.github/workflows/build.yml
micgor32: done

.github/workflows/build.yml
micgor32: done

walterchris (approved):

